PR #15, opened 2022-02-05
zigs-utxo: add new contract: zigs but using a UTXO-style model

hosted-fornet:
Logan suggested today to add a UTXO-style version of the zigs contract. The code here as of faf4fe1 is an initial, WIP implementation.

The main change relative to the current zigs contract is to what `rice` are doing. In the current zigs contract, there is one `rice` associated with the contract that has a map containing all user balances.

In contrast, here, there are many `rice`s, each one a "UTXO". A spend completely consumes a UTXO/`rice`.

FYI @tacryt-socryp @dr-frmr


dr-frmr:
Is it alright if we close this PR? I want to get rid of the dr/new-chick-types branch but it's gonna close this if I do. We can save the UTXO token contract for our contract library.

hosted-fornet:
closed as requested

PR #23, opened 2022-03-11
dao-agent step 2: DAO groups with: finer permissions, nesting, app integrations

hosted-fornet:
We want a custom groups-like backend for DAOs. The plan for the initial version is copied from [here](web+urbitgraph://group/~datwet/uqdev/graph/~datwet/open-questions-1325/170141184505530216236609377728995524608) below:

## The goal

The initial version of the DAO group agent will provide:

1. Nesting of DAO groups within DAO groups (already works, landscape just doesn't know how to render it).
2. Ability to change hosts of DAO groups/graphs ([some of this work is done](https://github.com/uqbar-dao/ziggurat/pull/20)).
3. More flexible permissioning, specifically enabling per-resource permissions within a DAO group (main focus of this post).
4. Flexibility to allow app integrations, such as calendar.

## The plan

The source of truth for all data about membership, permissions, etc is the chain. Users keep a local copy of some of this data in their DAO group store, see above. The off-chain copy will be synced with the on-chain copy by the DAO group store, which will subscribe to a chain indexer.

The plan is to initially write a minimal DAO group store (open to alternate names). It will store the off-chain copy of the DAO data and monitor the chain for updates.

One possible method for altering chain state is:

* DAO group store provides methods for modification of the off-chain state.
* If off-chain state is modified by a user with permissions to change the chain state, send tx to the chain to update the chain state.

We'll need very slightly modified graph push/pull hooks since they read from group store for permissioning; instead our versions will read from the DAO group store.

4d4cb23 adds a dumb database currently titled "dao-group-store" (name to be changed).


app/dao-graph-push-hook.hoon
```diff
@@ -0,0 +1,462 @@
...
+  ^-  permission-level:store
+  ?:  is-admin
+    admin.permissions
+  ?:  is-writer
+    writer.permissions
+  reader.permissions
+::
+++  get-permissions-variation
+:: ++  get-roles-writers-variation
+::   ~/  %get-roles-writers-variation  ::  TODO: write jet?
```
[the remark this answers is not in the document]
hosted-fornet: Whoops. Uncommented in ee6fa10. Sounds good, it would be good to learn more about this.

app/dao-group-store.hoon
```diff
@@ -0,0 +1,462 @@
...
+        ?~  dao-group=(peek-dao-group rid)  ~
+        ?~  wi=(watch-indexer rid dao-id.u.dao-group)  ~
+        ~[u.wi]
+      ::
+          %fact
+        =+  !<(=update:uqbar-indexer q.cage.sign)
+        ?>  ?=(%rice -.update)
+        =/  new-dao-group=dao-group:store
+          ;;(dao-group:store rice.update)  :: TODO: instead send diff?
+        ?.  .=  0  %~  wyt  in
```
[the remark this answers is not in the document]
hosted-fornet: Done in 4099239

app/dao-group-store.hoon
```diff
@@ -0,0 +1,462 @@
...
+            (~(get ju members.new-dao-group) our.bowl)
+          :-  ~
+          %=  this
+            dao-groups  (~(put by dao-groups) rid new-dao-group)  :: TODO: instead walk through and make minimal change to existing structure?
```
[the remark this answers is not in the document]
hosted-fornet: I recalled you had said this but didn't know how to implement it. Thinking about it more now, one method that comes to mind is sending a list of cards: the updates to do. For example, if we removed a user, we would send something like

```
~[[%remove-member [~zod %rug-dao] ~bus]]
```

and maybe also a hash of the expected final state.

app/dao-group-store.hoon
```diff
@@ -0,0 +1,462 @@
...
+    =/  pairs=(list (pair address:store role:store))
+      (make-noun-role-pairs address roles)
+    %+  %~  put  by  permissions
+      name
+    %-  %~  gas  ju
+      ?~  permission
+        *(jug address:store role:store)
+      u.permission
+    pairs
+  ::  TODO: use polymorphism?
```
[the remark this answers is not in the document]
hosted-fornet: Old comment -- sorry about that, removed in 3d9fd7c. It was from before I fixed a bug when `+remove-permissions-helper` and `+remove-roles-helper` looked like they might be able to be replaced by a single wet gate.

sur/dao-group-store.hoon
```diff
@@ -0,0 +1,48 @@
...
+      [%add-member =resource:r roles=(set role) =id =ship]
+      [%remove-member =resource:r =ship]
+      [%add-permissions =resource:r name=@tas =address roles=(set role)]
+      [%remove-permissions =resource:r name=@tas =address roles=(set role)]
+      [%add-subdao =resource:r subdao-id=id]
+      [%remove-subdao =resource:r subdao-id=id]
+      [%add-roles =resource:r roles=(set role) =ship]
+      [%remove-roles =resource:r roles=(set role) =ship]
+  ==
++$  update
```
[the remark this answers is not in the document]
hosted-fornet: Good point, there isn't much benefit of separating these as is. I think I had previously had all the `modify` require `%write` privileges to the DAO, but realized `%remove-group` should be runnable by anyone, and so the distinction broke down.

Combined these types in 1ba11b9

hosted-fornet:
Some random thoughts and TODOs:

1. DAO owners can just be members with `%write` permissions on the DAO id. The upside of this is you don't need a special data structure (like `owners=(set id)`). The downside is it is harder to read off at a glance who the owners are. Perhaps make a utility function to print off the owners as a set? Or more generally, the users with a given role/permission? Another possible downside: it may be more costly to verify DAO owners on the chain when sending changes to the DAO.
2. I think it makes sense to split this into two agents: one a true DAO agent and one something like `%dao-comms`. The DAO agent's job is to store a local copy of the chain state of whatever DAOs the user wants and to keep them updated by subscribing to an indexer. The `%dao-comms` agent job is to read from the data stored in the DAO agent and then do all the permissioning stuff with graphs/apps as necessary. The reason to split these is because other apps will need to make use of the DAO agent's local DAO data. For example, it makes more sense to have a %dao-treasury app talking to the DAO agent than a combined DAO/%dao-comms agent.

Still TODO:

I. Allow owner to push changes from local store to the chain. I think it makes sense to have a poke that can only be called by an owner, like `%send-to-chain` that will send changes to be made to chain.
II. Update existing DAO contract to work with these data structures.
III. Split out %dao and %dao-comms, if desired.

hosted-fornet:
@tacryt-socryp please take a look when you've got a chance, I got host changing working for DAOs in 6ba4860.

There is still a lot to improve here, but this code is now working and this PR has been open for a while. Below is how to test creating a DAO on-chain, mirroring its state off-chain on multiple ships, manipulating its state on-chain (with corresponding updates to off-chain state), and changing comms hosts.

My particular set up used 3 ships: ~zod, the original host, ~nec, the new host, and ~bus, a pleb member. The chain is somewhat buggy as of the code inthis PR, which is part of the reason I want to get this merged so I can start developing on top of @dr-frmr s more recent and stable chain. As such, if ~zod and ~nec get out of sync, you may have to restart the setup process.

Start each ship and copy the code here into a %zig desk. After installing:

```
::  on ~zod
=pubkey 0x2.e3c1.d19b.fd3e.43aa.319c.b816.1c89.37fb.b246.3a65.f84d.8562.155d.6181.8113.c85b
=zigs-id 0x10b.4ca5.fb93.480b.a0d7.c168.0f3e.6d43
=dao-id 0xef44.5e1e.2113.c21d.7560.c831.6056.d984
=nec-pubkey 0x3.4cdd.5f53.b551.e62f.2238.6eb3.8abd.3e91.a546.fad3.2940.ff2d.c316.50dd.8d38.e609
:wallet &zig-wallet-poke [%populate 0xbeef]
:ziggurat &zig-chain-poke [%set-addr pubkey]

:ziggurat|start-testnet now

::  start-testnet will spin for ~30s; while spinning, can run the following on ~nec:
=pubkey 0x3.4cdd.5f53.b551.e62f.2238.6eb3.8abd.3e91.a546.fad3.2940.ff2d.c316.50dd.8d38.e609
=zigs-id 0x10b.4ca5.fb93.480b.a0d7.c168.0f3e.6d43
=dao-id 0xef44.5e1e.2113.c21d.7560.c831.6056.d984
:wallet &zig-wallet-poke [%populate 0xdead]
:ziggurat &zig-chain-poke [%set-addr pubkey]

::  and on ~bus:
=pubkey 0x3.9452.264c.57a5.1b54.d380.70b0.7e0c.934d.15c0.6692.fa9c.7f35.eaf9.eb52.b925.1b7d
=zigs-id 0x10b.4ca5.fb93.480b.a0d7.c168.0f3e.6d43
=dao-id 0xef44.5e1e.2113.c21d.7560.c831.6056.d984
:wallet &zig-wallet-poke [%populate 0xcafe]
:ziggurat &zig-chain-poke [%set-addr pubkey]

::  once spinning of start-testnet is done on ~zod, run on ~zod:
:uqbar-indexer &set-chain-source [our %ziggurat]
:sequencer|init 1
:dao &set-indexer [our %uqbar-indexer]
-zig!create-dao-comms [[~zod %uqbar-dao] `@`'uqbar-dao' 'Uqbar DAO']

::  init will spin for ~30s. once that is done, to join the chain as a validator and watch for DAO updates run on ~nec:
:ziggurat &zig-chain-poke [%start %validator ~ validators=(silt ~[~zod ~nec]) [~ ~]]
:uqbar-indexer &set-chain-source [our %ziggurat]
:dao &set-indexer [our %uqbar-indexer]
-zig!create-dao-comms [[~zod %uqbar-dao] `@`'uqbar-dao' 'Uqbar DAO']

::  ~bus will just be listening to ~zod's ziggurat rather than validating; run on ~bus:
:uqbar-indexer &set-chain-source [~zod %ziggurat]
:dao &set-indexer [our %uqbar-indexer]
-zig!create-dao-comms [[~zod %uqbar-dao] `@`'uqbar-dao' 'Uqbar DAO']

::  set up graphs on ~zod:
-zig!create-dao-graph [[~zod %uqbar-dao] [~zod %dao-chat] 'dao chat' 'place to chat about the dao' %chat]
-zig!create-dao-graph [[~zod %uqbar-dao] [~zod %dao-links] 'dao links' 'place to share links relevant to the dao' %link]
=g -build-file /=landscape=/sur/graph-store/hoon
:dao-graph-push-hook &graph-update-3 [now %add-nodes [~zod %dao-chat] (~(put by *(map index:g node:g)) /1 [[%& our /1 now ~[[%text 'hi']] ~ ~] [%empty ~]])]
:dao-graph-push-hook &graph-update-3 [now %add-nodes [~zod %dao-links] (~(put by *(map index:g node:g)) /1 [[%& our /1 now ~[[%text 'le mems'] [%url 'https://urbit.org']] ~ ~] [%empty ~]])]

::  the metadata-store on ~nec and ~bus will show these graphs, but to get them, must run the following on ~nec and ~bus:
:dao-graph-pull-hook &pull-hook-action [%add ~zod [~zod %dao-chat]]
:dao-graph-pull-hook &pull-hook-action [%add ~zod [~zod %dao-links]]

::  finally, run each of these DAO actions in a single block, sending from ~zod:
::  the first two add proposals which should be reflected in the DAO state `proposal` map;
::  the second two pass each of the first two proposals
::  once the second vote is processed, ~nec will be the new %comms-host and there should be some scroll
::  in the Dojos of each ship. inspecting the state should show a proper transfer of push/pull hooks,
::  metadata-store, and graph-store.
::  posts in [~nec %dao-chat] should now be possible (but not [~nec %dao-links] because no one has %write permissions)
=remove-zod-role-host-update [%remove-roles dao-id (~(put in *(set @tas)) %comms-host) pubkey]
=add-nec-role-host-update [%add-roles dao-id (~(put in *(set @tas)) %comms-host) nec-pubkey]
-zig!send-dao-action [our [pubkey 1 zigs-id] [%propose dao-id remove-zod-role-host-update]]
-zig!send-dao-action [our [pubkey 2 zigs-id] [%propose dao-id add-nec-role-host-update]]
-zig!send-dao-action [our [pubkey 3 zigs-id] [%vote dao-id 0x1ee9.92c7]]
-zig!send-dao-action [our [pubkey 4 zigs-id] [%vote dao-id 0x17c2.5ecf]]
```

Tests on the DAO contract can be run by uncommenting line 10 in `lib/zig/contracts/dao.hoon` and running

```
-test /=zig=/tests/contracts/dao/hoon ~
```

Outstanding issues:
1. There is some error scroll when host is changed.
2. The hooks to the original host are not all severed as expected.

Desired additional features:
1. Change proposals from a single action to a list of actions, to be executed in order if passed.
2. Change off-chain DAO agent to make minmial, semantic changes to DAO state when receiving updates from chain, rather than overwriting as it currently does. One possible way is to make a diff between old and new DAO state capture the proposal that passes, and run those actions.

hosted-fornet:
Currently merging with master (which now includes the wallet branch), so please delay review until I've got those commits sent.

hosted-fornet:
ca256ea updates DAO, indexer, wallet, and ziggurat codes to play nice together post-merge.

Some notes:

I've updated the instructions for testing [above](https://github.com/uqbar-dao/ziggurat/pull/23#issuecomment-1129611339) to reflect updated keys and a few other new things. Two things worth calling out here:

1. In order to join an ongoing chain as a validator, you must first get at least one block from that chain in your indexer, e.g. `:uqbar-indexer &set-chain-source [~zod %ziggurat]`. The instructions above reflect this.
2. There appears to be a bug where you need to call `:sequencer|init-dao 1` (or `init` as well, I assume) twice because in the first call, the wallet does not have the necessary data populated yet, and so it fails, but populates the data, which the second call can then use. @dr-frmr

hosted-fornet:
Changes are done on this PR. Going to merge changes here to chained PRs, but this PR should be good to review again.

hosted-fornet:
> 2. There appears to be a bug where you need to call `:sequencer|init-dao 1` (or `init` as well, I assume) twice because in the first call, the wallet does not have the necessary data populated yet, and so it fails, but populates the data, which the second call can then use. @dr-frmr

I think this is related to when the `:wallet ... %populate ... ` is called and whether the indexer has blocks in it yet. I've updated the instructions above to reflect this. The solution is to run `:wallet ... %populate ...` after the chain has been started and the indexer has received at least one block, but before running the sequencer.

Note this gets extra broken in PRs down the chain because the wallet subscribes to the indexer. Here we send the data the wallet needs every block, but down the chain, we only send it on initial subscription (which, if we don't have the data, means we send nothing) and then upon change of the data, which will never happen since we can't start the sequencer.

app/dao-metadata-pull-hook.hoon
```diff
@@ -0,0 +1,338 @@
...
+        =+  !<(=update:contact q.cage.sign)
+        ?+  -.update  `state
+            %add
+          =/  missing=(set resource)
+            (check-contact contact.update)
+          =.  pending
+            (~(uni in pending) missing)
+          :_  state
+          (get-many-previews missing)
+
```
tacryt-socryp, line 124: no newlines

app/dao-metadata-pull-hook.hoon
```diff
@@ -0,0 +1,338 @@
...
+::
+++  on-save  !>(state)
+++  on-load  
+  |=  =vase
+  =+  !<(old=versioned-state vase)
+  |^  
+  ?-  -.old
+    %2  `this(state old)
+    ::
+      %1
```
tacryt-socryp, line 211: Are the upgrades necessary or can we just throw away state for prior iterations?

app/dao-metadata-pull-hook.hoon
```diff
@@ -0,0 +1,338 @@
...
+  ::
+      [%preview @ @ @ ~]
+    =/  rid=resource
+      (de-path:resource t.path)
+    =/  prev=(unit group-preview:metadata)
+      ?^  (peek-metadatum:met %groups rid)  
+        (some (get-preview:met rid))
+      (~(get by previews) rid)
+    ?~  prev
+      :_  this(pending (~(put in pending) rid))
```
tacryt-socryp, line 309: Preferably :- given that the this statement is longer.

app/dao-metadata-push-hook.hoon
```diff
@@ -0,0 +1,198 @@
...
++$  state-zero  [%0 ~]
+::
++$  versioned-state
+  $%  state-null
+      state-zero
+  ==
+--
+::
+::
+=+
```
tacryt-socryp, line 34: A lot of boilerplate just to avoid repetition of the watch-groups call, no?

app/uqbar-indexer.hoon
```diff
@@ -222,13 +254,16 @@
...
-      :^  ~  ~  %noun
-      !>  ^-  (unit update:uqbar-indexer)
-      %-  combine-update-sets
-      ;;  %-  list
-        %-  unit
-        [%egg eggs=(set [egg-location:uqbar-indexer egg:smart])]
-      ~[from to]
+      =/  up=(unit update:uqbar-indexer)
+        %-  combine-update-sets
+        ;;  %-  list
```
tacryt-socryp, line 259: Why do you need to ;; here?

app/uqbar-indexer.hoon
```diff
@@ -241,13 +276,105 @@
         (serve-update %from hash)
       =/  to=(unit update:uqbar-indexer)
         (serve-update %to hash)
+      =/  up=(unit update:uqbar-indexer)
+        %-  combine-update-sets
+        ;;  %-  list
```
tacryt-socryp, line 281: Same

lib/uqbar-indexer.hoon
```diff
@@ -0,0 +1,497 @@
...
+    :-  ~
+    %.  jon
+    %-  ot
+    :^    [%height ni]
+        [%signature signature]
+      [%chunks chunks]
+    ~
+  ::
+  ++  town
+    |=  jon=json
```
tacryt-socryp, line 476: When doing this pattern of a JSON parser, it's neater and has one fewer gate call to instead do a cast as follows:
`^-  $-(json town:smart)` then write the parser statement (in this case starting with ot) inline. No need for the %. or the |= when doing this pattern.

app/dao.hoon
```diff
@@ -282,25 +309,45 @@
...
+      :+  (remove-push-hook 'graph')
+        (remove-push-hook 'metadata')
+      ~
+      ::
+      ++  remove-push-hook
+        |=  app-infix=@ta
+        ^-  card
+        =/  full-app-name=@tas
+          (rap 3 %dao '-' app-infix '-' %push-hook ~)
+        %^  %~  poke-our  pass:io  (weld remove-wire /[app-infix])
```
tacryt-socryp, line 326: Surely the %~ portion of this statement would be cleaner in wideform given that it's still on a single line

tacryt-socryp (commented):
Overall looks great. One thing I'd recommend before merging into master is renaming uqbar-indexer to indexer. Should be a simple find and replace, and the succinctness is nicer. Overall just some style recommendations and questions, architecture is great.

app/uqbar-indexer.hoon
hosted-fornet: As currently written, the `;;` cast is required because `+combine-update-sets` expects a `(list (unit [%egg ...]))`, but the input is a `(list (unit update:uqbar-indexer))`. The compiler complains without the `;;` cast.

An alternative way to write this might be to have `+combine-update-sets` accept a `(list (unit update:uqbar-indexer))` and then crash on/ignore non-`%egg` updates. Could make this change if you prefer this pattern.

[A few PRs down the line](https://github.com/uqbar-dao/ziggurat/pull/47/files#diff-e638653ea0ee7afc193943cc48078a89ae6f55d173e0602cabad8569dad5cb6cR767), `+combine-update-sets` is renamed to the more correct `+combine-egg-updates`.

hosted-fornet:
> Overall looks great. One thing I'd recommend before merging into master is renaming uqbar-indexer to indexer. Should be a simple find and replace, and the succinctness is nicer. Overall just some style recommendations and questions, architecture is great.

Thanks for the review. Would you mind if I tack on a PR at the end of one of these PR chains to make the `uqbar-indexer` -> `indexer` rename rather than making it here? Would avoid having to propagate the change through the chain.

app/uqbar-indexer.hoon
tacryt-socryp: Ah. It would be more performant to make a helper function that `+turn`'s over the list in question and asserts with `?>` that the update is of the `%egg` subtype.

app/dao.hoon
hosted-fornet: Good point. Done in f40712d

lib/uqbar-indexer.hoon
hosted-fornet: Neat. Done in 304d02b

app/dao-metadata-push-hook.hoon
hosted-fornet: Removed in 2e71321

app/dao-metadata-pull-hook.hoon
hosted-fornet: Removed this historical state code in 0fab6b5, as well as addressing other `dao-metadata-pull-hook.hoon` comments.

app/uqbar-indexer.hoon
hosted-fornet: Changed in 8da50a4 to avoid need to `;;`.

PR #32, opened 2022-04-08
zink: restructured file to allow threading through budget

tacryt-socryp:
This is a draft.

We're doing basic gas budgeting, and I've added in the helper methods `+frag` and `+edit` from `+bink`. We'll need to do more work to remove crash sites, update the types to reflect that crashing is not allowed, build wrappers akin to `+mule`, and ensure we're generating hashes for the `+frag` and `+edit` ops.


martyr-binbex (commented):
We discussed this PR in a call and it looks great.

PR #40, opened 2022-04-26
Wallet agent (and miscellaneous improvements to smart contracts)

dr-frmr:
This PR includes a huge number of changes to all of the core agents in this repo. It also contains an outdated version of @hosted-fornet's DAO agent and indexer agent, so those should be merged into master before this, and I can resolve any conflicts here.

Included here is the Uqbar Wallet agent, which has a matching app interface in the wallet-ui repo. The wallet has a number of pokes and scrys available which are included in the updated readme here.

The sequencer and ziggurat agents have been upgraded to process all transactions through the wallet agent. They've also undergone a number of structural changes in order to make the chain more robust -- I've been running 1/2/3/4-ship testnets regularly with this code in order to hunt down edge cases and get the blockchain state machine ready to run in a persistent manner across a fleet of moons. The ziggurat, sequencer, wallet, and indexer agents are all intertwined, and instructions in the readme show how to start up a chain or join an existing one. (Ships running ziggurat/sequencer don't necessarily need to run an indexer, but in order to first join the chain, they need to submit a transactions with their wallet, which currently *does* need a local indexer to function).

Uqbar blocks now have a height. This is a separate number to slots within epochs, and is simply an integer incremented on every successful block. This can now finally be used inside contracts and elsewhere.

Remote scry will prompt another wave of changes to all of these agents. We'll probably be able to rewrite many parts of the machinery to be more efficient, with (I'm guessing) less and cleaner code.

There are also changes to mill.hoon, the contract execution engine, included in this PR. Mill is now capable of properly handling continuation calls, while the `event` arm of contracts has been removed. Contract reads are now (theoretically) enabled via scrys in sequencer.

Mill also supports ETH addresses now. If a transaction `egg` contains an `eth-hash` it verifies signature based on that. This allows hardware wallet support.

Contract write results now contain `crow`, a list of arbitrary json objects which can be used as event announcements, similar to emitted events in solidity.

The uHoon standard library has been expanded to include the json type from lull.hoon and +enjs arms from zuse.hoon, so contracts can now define the noun and json data formats of rice they have provenance over. All existing contracts have been updated to fit this standard.


dr-frmr:
Noteworthy bug, TODO for myself: occasionally a ship will receive a chunk before a block. Need to add a delay in sequencer.hoon that makes chunk-production occur partway through slot, as with block production.

dr-frmr:
Long term, there will have to be some kind of mechanism for chunk producers (sequencers) to recognize a non-responsive block producer and intelligently send their chunks to someone else.

dr-frmr:
> Noteworthy bug, TODO for myself: occasionally a ship will receive a chunk before a block. Need to add a delay in sequencer.hoon that makes chunk-production occur partway through slot, as with block production. Happens wonderfully infrequently...

Saving this for future PR. Final TODO before merging this is to bring in latest indexer changes and accommodate new indexer-update format.

PR #41, opened 2022-05-07
indexer: json support

hosted-fornet:
WIP


tacryt-socryp:
Styling and structure looks really good.

hosted-fornet:
Thanks!

As of 1cf5fd7 JSON scries are working according to my testing. In addition to these changes, I'll also be adding:
1. A way to get `x` most recent block numbers. We could compute this on the fly or maintain a running list to make this an indexed query. I think the second makes the most sense considering this is an indexer.
2. Time/block-number bounded transaction histories for accounts. To do right, this will touch `sequencer` code, so I need to be careful to avoid merge conflicts.

hosted-fornet:
I have not added 2 in the above comment. I'm going to merge this now. I'd like to finish work on the DAO agent branch and get that merged to master before trying to work on 2, so that I can avoid merge conflicts by branching off of https://github.com/uqbar-dao/ziggurat/pull/40

We'll see if that is possible based on @willbach 's requirements.

PR #43, opened 2022-05-19
indexer: send to subscribers only initially and on diffs

hosted-fornet:
Before this PR, a subscriber would be sent, e.g., the requested grain with every block, because the grain was contained in every block.

In this PR, a subscriber is sent, e.g., the requested grain upon subscription and then only when there are changes to the grain. This should greatly reduce the amount of communication the indexer must do to subscribers, since most of the chain state is expected to be static in a given block.


app/uqbar-indexer.hoon
```diff
@@ -154,14 +156,71 @@
...
+      ?~  newest-epoch=(pry:poc:zig epochs)  ~
+      ?~  newest-slot=(pry:sot:zig slots.val.u.newest-epoch)
+        ~
+      =*  newest-block  q.val.u.newest-slot
+      ?~  newest-block  ~
+      =*  newest-chunks  chunks.u.newest-block
+      =/  newest-chunk  (~(get by newest-chunks) town-id)
+      ?~  newest-chunk  ~
+      :_  ~
+      %+  fact:io
```
tacryt-socryp, line 171: Neater to write as

```
%- fact:io
:_  ~
```

then the statement

app/uqbar-indexer.hoon
```diff
@@ -154,14 +156,71 @@
...
+        (serve-previous-update %to payload)
+      =/  update=(unit update:uqbar-indexer)
+        %-  combine-update-sets
+        ;;  %-  list
+          %-  unit
+          [%egg eggs=(set [egg-location:uqbar-indexer egg:smart])]
+        ~[from-update to-update]
+      ?~  update  ~
+      :_  ~
+      %+  fact:io
```
tacryt-socryp, line 196: Same

app/uqbar-indexer.hoon
```diff
@@ -154,14 +156,71 @@
...
+      ~
+    ::
+        ?([%grain @ ~] [%holder @ ~] [%lord @ ~])
+      :_  this
+      =/  query-type=?(%grain %holder %lord)  i.path
+      =/  payload=@ux  (slav %ux i.t.path)
+      ?~  update=(serve-previous-update query-type payload)
+        ~
+      :_  ~
+      %+  fact:io
```
tacryt-socryp, line 208: Same

app/uqbar-indexer.hoon
```diff
@@ -154,14 +156,71 @@
...
+        !>(`update:uqbar-indexer`u.update)
+      ~
+    ::
+        [%slot ~]
+      :_  this
+      ?~  newest-epoch=(pry:poc:zig epochs)  ~
+      ?~  newest-slot=(pry:sot:zig slots.val.u.newest-epoch)
+        ~
+      :_  ~
+      %+  fact:io
```
tacryt-socryp, line 219: Same

app/uqbar-indexer.hoon
```diff
@@ -92,6 +92,8 @@
   $:  =epochs:zig
       num-recent-headers=@ud
       recent-headers=(list [epoch-num=@ud =block-header:zig])
+      $=  serve-previous-update
+    $-([=query-type:uqbar-indexer =query-payload:uqbar-indexer] (unit update:uqbar-indexer))
```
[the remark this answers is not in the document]
hosted-fornet: I need to be able to have access to an index of only the most recent block. I use this for two things:
1. To determine if items have changed to determine whether to send an update to subscribers.
2. To send initial update to new subscribers: the state of the item they are subscribing to.

I was already computing a "most recent block index" to create cards to send to subscribers (see, e.g., [here](https://github.com/uqbar-dao/ziggurat/blob/a788964a89320e384a50e8653760efc8f115f4e2/app/uqbar-indexer.hoon#L354-L407)), so it was easy to just store this. Then when a new block comes in, compute the newest "most recent block index", and it is easy to compare diffs, etc.

An alternative to this strategy might be to change the index `jug`s to `jar`s. Then if I query for a grain, say, the most recently added item in the list is the most recently indexed state of the grain.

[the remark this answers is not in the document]
tacryt-socryp: I don't think there's any reason for you to store a function in state this way and it can make state upgrades a little funky as well as making `+on-load` take significantly longer.

app/uqbar-indexer.hoon
hosted-fornet: Nice. Done here and below in 9cfce75

app/uqbar-indexer.hoon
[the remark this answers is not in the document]
hosted-fornet: Okay, I changed this in 6ed1013. Where we previously stored the function, now we store the data used to create the function and create it as needed.

tacryt-socryp (approved):
LGTM

app/uqbar-indexer.hoon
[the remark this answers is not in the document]
tacryt-socryp: Perfect

PR #44, opened 2022-05-20
indexer: use `%ziggurat` catch up to get initial blocks

hosted-fornet:
Previously, the indexer started indexing from whatever the current block at indexer start time was.

This PR uses the `%ziggurat` `epoch-catchup` endpoint to get blocks since the beginning of time. That endpoint just gives all the `epochs`, which we recurse through and consume one-by-one, adding each to the index.


app/uqbar-indexer.hoon
```diff
@@ -470,30 +484,57 @@
...
+            $(slots rest.slot, cards new-cards, state new-state)
+        $(epochs rest.epoch, cards new-cards, state new-state)
       ::
           %indexer-block
-        ?~  blk.update  `state  :: TODO: log block header?
-        =*  epoch-num   epoch-num.update
-        =*  block-num   num.header.update
-        =/  new-slot=slot:zig  [header.update blk.update]
+        %+  consume-slot
+        epoch-num.update  [header.update blk.update]
```
tacryt-socryp, line 507: epoch-num.update should be on prior line

app/uqbar-indexer.hoon
```diff
@@ -600,15 +638,25 @@
...
+  /epochs-catchup
+::
 ++  chain-update-wire
   ^-  wire
   /chain-update
 ::
+++  get-epoch-catchup
+  |=  d=dock
+  ^-  card
+  %+  %~  watch  pass:io
```
tacryt-socryp, line 652: `%~` wideform

tacryt-socryp (approved):

app/uqbar-indexer.hoon
hosted-fornet: fixed in 8111d27

app/uqbar-indexer.hoon
hosted-fornet: fixed in 8111d27

PR #45, opened 2022-05-20
dao: use list of updates in proposals

hosted-fornet:
Each proposal previously contained one update to run if the proposal passed. Now, each proposal contains a list of updates to be run in order if the proposal passes. This allows, e.g., a host change to be accomplished in one, rather than two or more, proposals.

The advantage here is two-fold:
1. Reduce cognitive load on DAO participants by bundling connected actions together.
2. Reduce the cost to DAO participants by reducing the number of votes required by bundling actions.


hosted-fornet:
Similar instructions for testing as [here](https://github.com/uqbar-dao/ziggurat/pull/23#issuecomment-1129611339), but use lists in proposals, e.g.

```
=change-comms-host-updates ~[[%remove-roles dao-id (~(put in *(set @tas)) %comms-host) pubkey] [%add-roles dao-id (~(put in *(set @tas)) %comms-host) nec-pubkey]]
-zig!send-dao-action [our [pubkey 1 zigs-id] [%propose dao-id change-comms-host-updates]]
-zig!send-dao-action [our [pubkey 2 zigs-id] [%vote dao-id 0x5f12.f621]]
```

tests/contracts/dao.hoon
```diff
@@ -104,6 +107,19 @@
         make-placeholder-dao-comms-rid
       (~(put in *(set role:d)) %jannie)
     ::
+    ++  make-add-comms-host-permissions-and-role-updates
+      ^-  (list on-chain-update:d)
+      :+  :-  %add-permissions
```
tacryt-socryp, line 112: Desireable to place long expressions as close to the left side of the page (lack of indents) as possible. I recommend instead of `:+` using `:_  ~  :_  ~`

tests/contracts/dao.hoon
```diff
@@ -611,6 +642,72 @@
   ::
   --
 ::
+++  test-propose-valid-proposal-multiple-updates
+  |^
+  =/  expected-chick=chick:smart  make-expected-chick
+  =/  =cart:smart                 make-test-cart
+  =/  =zygote:smart               make-zygote
+  =+  [is-success chick]=(mule |.((~(write cont cart) zygote)))
+  ;:  weld
```
tacryt-socryp, line 651: Weld already takes only two arguments, can use %+. No need to use the n-ary override rune

tests/contracts/dao.hoon
```diff
@@ -611,6 +642,72 @@
...
+      !>  chick
+  ==
+  ::
+  ++  make-expected-chick
+    ^-  chick:smart
+    =/  =dao:d  megacorp-dao
+    =/  updates=(list on-chain-update:d)
+      make-on-chain-updates
+    :-  %&
+    :+  %+  make-id-grain-map  megacorp-dao-salt
```
tacryt-socryp, line 666: Better with colcabs like the other

tests/contracts/dao.hoon
```diff
@@ -1096,4 +1223,56 @@
...
   --
 ::
+++  test-execute-final-vote-multiple-updates
+  |^
+  =/  expected-chick=chick:smart  make-expected-chick
+  =/  =cart:smart                 make-test-cart
+  =/  =zygote:smart
+    (make-megacorp-ceo-vote-zygote make-proposal-id)
+  =+  [is-success chick]=(mule |.((~(write cont cart) zygote)))
+  ;:  weld
```
tacryt-socryp, line 1233: %+

tests/contracts/dao.hoon
```diff
@@ -1096,4 +1223,56 @@
...
+    ^-  id:smart
+    %-  mug
+    %-  jam
+    make-add-comms-host-permissions-and-role-updates
+  ::
+  ++  make-expected-chick
+    ^-  chick:smart
+    =/  =dao:d  megacorp-dao
+    :-  %&
+    :+  %+  make-id-grain-map  megacorp-dao-salt
```
tacryt-socryp, line 1252: Better with colcabs

tests/contracts/dao.hoon
hosted-fornet: Changed in 3c3f50f; was following style as seen in Sigilante's testing guide [here](https://medium.com/dcspark/writing-robust-hoon-a-guide-to-urbit-unit-testing-82b2631fe20a).

I kept the non-standard indentation of the `+expect-eq` args, but can change that too if preferred.

tests/contracts/dao.hoon
hosted-fornet: I think I'm missing something here. Unless I'm mistaken, a `:+  'foo'  'bar  'baz'` is different from a `:_  'baz'  :_  'bar'  'foo'`:

```
:+  'foo'  'bar  'baz'  ::  ['foo' ['bar' 'baz']]

:_  'baz'  :_  'bar'  'foo'`  ::  [['foo' 'bar'] 'baz']
```

so I don't understand how to get equivalent cell structures using two colcabs.

tacryt-socryp: yeah you're totally right

tacryt-socryp (approved):
LGTM

PR #47, opened 2022-05-24
indexer: add %hash query over eggs, grains, and blocks

hosted-fornet:
Enable the `%hash` scry endpoint.

Some internal and external-facing changes in order to do so.

Internally changed `from-index` and `to-index` to second-order, i.e. store the hash for an egg rather than a pointer to the location in a block. This makes it easy to get the egg hash which is important for...

Externally changed `%egg` update data to hold a map from egg hash to data. Also changed `%slot` update to match this pattern. This should make life easier for requesters to access data and standardizes the update data type somewhat.

Added the `%hash` update data type, which has maps for each of `eggs`, `grains`, and `slots`. Keeping the maps separate (as opposed to returning a type something like `[%hash items=(map id ?(egg grain slot))]`) makes access easier and also reduces the amount of type-wrangling we have to do in the indexer.

Also fixed some bugs in the JSON (de)parsing.


app/uqbar-indexer.hoon
```diff
@@ -712,28 +737,73 @@
...
-  `[%egg combined]
+    ~(tap by eggs.u.update)
+  combined
+::
+++  combine-grain-updates-to-map
+  |=  updates=(list (unit update:uqbar-indexer))
+  ^-  (map id:smart [town-location:uqbar-indexer grain:smart])
+  ?~  updates  ~
+  =/  combined=(map id:smart [town-location:uqbar-indexer grain:smart])
+    %-  %~  gas  by  *(map id:smart [town-location:uqbar-indexer grain:smart])
```
tacryt-socryp, line 768: put the `%~` part of this statement in wideform when it fits on a single line

app/uqbar-indexer.hoon
```diff
@@ -779,10 +849,10 @@
     ?.  ?=(%egg -.u.two)  %.n
     =/  two-eggs=(set egg:smart)
       %-  %~  gas  in  *(set egg:smart)
-      %+  turn  ~(tap in eggs.u.two)
+      %+  turn  ~(val by eggs.u.two)
       |=  [egg-location:uqbar-indexer =egg:smart]
       egg
-    %-  %~  all  in  eggs.u.one
+    %-  %~  all  by  eggs.u.one
```
tacryt-socryp, line 855: I recommend `%~` in wideform since it fits on one line

app/uqbar-indexer.hoon
```diff
@@ -799,7 +869,14 @@
   ::
       %slot
     ?.  ?=(%slot -.u.two)  %.n
-    =(slot.u.one slot.u.two)
+    =/  two-slots=(set slot:zig)
+      %-  %~  gas  in  *(set slot:zig)
+      %+  turn  ~(val by slots.u.two)
+      |=  [block-location:uqbar-indexer =slot:zig]
+      slot
+    %-  %~  all  by  slots.u.one
```
tacryt-socryp, line 877: same

app/uqbar-indexer.hoon
```diff
@@ -991,27 +1077,51 @@
...
-          grains
-        (~(uni by grains.u.out) grains.u.next-update)
+      ?+  -.u.out  !!
+      ::
+          %egg
+        ?>  ?=(%egg -.u.next-update)
+        %=  u.out
+            eggs
+          (~(uni by eggs.u.out) eggs.u.next-update)
+        ::
```
tacryt-socryp, line 1115: unnecessary ::

app/uqbar-indexer.hoon
```diff
@@ -991,27 +1077,51 @@
...
+        ==
+      ::
+          %grain
+        ?>  ?=(%grain -.u.next-update)
+        %=  u.out
+            grains
+          (~(uni by grains.u.out) grains.u.next-update)
+        ::
```
tacryt-socryp, line 1123: same

lib/uqbar-indexer.hoon
```diff
@@ -266,9 +296,19 @@
...
         [%grain grains]
-        [%slot slot]
+        [%hash (ot ~[[%eggs eggs] [%grains grains] [%slots slots]])]
+        [%slot slots]
     ==
   ::
+  ++  block-location
+    |=  jon=json
+    ^-  block-location:ui
+    %.  jon
```
tacryt-socryp, line 306: can simplify this pattern as in the other PR

lib/uqbar-indexer.hoon
```diff
@@ -413,6 +453,16 @@
     :-  [%owns ids]
     ~
   ::
+  ++  slots
+    |=  jon=json
+    ^-  (map slot-id=id:smart [location=block-location:ui =slot:zig])
+    %.  jon
```
tacryt-socryp, line 459: same as other PR

tacryt-socryp (commented):
Overall good. Not particular to this PR, but I see a lot of mixed `(unit)` and crash semantics. I think you should make firm decisions about which parts of this program should or should not crash, and make those semantics consistent to match that desire. Right now, there are functions where you both return `~` on a unit and also crash on another type assertion. Pick one!

hosted-fornet:
9e82c82 addresses your specific comments. As for the more general comment on returning a null `unit` vs crashing, I've put this on my todo list and will address this in another PR unless you insist I work on it here.

PR #53, opened 2022-05-26
sequencer/ziggurat: fix `read-rice`/`read-wheat` callsites

ynx0:
Fixes faulty arguments being passed into `+read-rice` and `read-wheat` respectively in the `on-peek` arms of both agents that use it. The util arms expect the path to be without the leading `%x` that is present in the path from the scry.


dr-frmr:
Thank you for fixing this, merging in

PR #54, opened 2022-05-29
README testnet instructions overhaul. Update wallet to work with latest %indexer design

dr-frmr:
wallet: change subscription path for txs to match above changes

I was trying to figure out why the wallet wasn't getting subscription updates on the `id` path from indexer and figured this would be easier. I originally asked for this path, but am seeing now that it's probably better for all of us to separate out the two paths, for updates to transactions a given ID sent, and ones they've received. This change seems to work and wondering if this will make the indexer also happier, since I see that it's some extra work to combine the `to` and `from` stuff.


hosted-fornet:
I think this change makes sense for subscriptions. It probably still makes sense to have scry endpoints for `/id` and `/hash`: the alternative is that the frontend will have to send a number of queries and combine them (think the main search box on etherscan). It makes more sense to me to do the combination in the backend.

dr-frmr:
Yeah, that makes sense for searches. It doesn't seem like the `id` sub path was working, though -- original impetus for this was that I wasn't getting any updates on that path

hosted-fornet (approved):
Looks good, works well :+1: 

Shouldn't be addressed here, but we should discuss the custom transactions at some point. I was testing with the DAO contract and it didn't work because I need to pass in data structures like `set`s and `map`s. Not sure I am using it right. An example of my poke:

```
:wallet &zig-wallet-poke [%submit-custom from=0x3.e87b.0cbb.431d.0e8a.2ee2.ac42.d9da.cab8.063d.6bb6.2ff9.b2aa.e1b9.0f56.9c3f.3423 to=`@ux`'dao' town=1 gas=[1 1.000] args='[%propose 0xef44.5e1e.2113.c21d.7560.c831.6056.d984 [%remove-roles 0xef44.5e1e.2113.c21d.7560.c831.6056.d984 (~(put in *(set @tas)) %comms-host) 0x3.e87b.0cbb.431d.0e8a.2ee2.ac42.d9da.cab8.063d.6bb6.2ff9.b2aa.e1b9.0f56.9c3f.3423]]' my-grains=~ cont-grains=(silt ~[0xef44.5e1e.2113.c21d.7560.c831.6056.d984])]
```

dr-frmr:
@hosted-fornet, you are using it right. The issue as @ynx0 pointed out to me is that the custom arguments are getting `ream`ed against an empty hoon, which means items like maps in the standard library aren't parseable. We will certainly want to change this, and now it's just a question of what to `slap` it against -- either `zuse` or our own compiled thing.

tacryt-socryp:
I recommend compiling it against our custom hoon.hoon and smart.hoon. Ideally everything is compiled against the exact same subject, as that will maximize compatibility.

dr-frmr:
Ok, after feedback, going to merge this without fully fixing %submit-custom and save that work for another branch.

PR #58, opened 2022-05-31
Draft: Zink / Mill integration: the Zebra Shut the Door

tacryt-socryp:
mill/zink: implement generic +slam on doors called +shut, implement +zebra which is bounded zk equivalent of +mule.

Also implemented partial eval of +shut specifically for use in mill, this one is called +ajar.


dr-frmr (approved):
Do you want me to try loading in a real contract on this and running it, or are we good?

dr-frmr:
Ok sounds great, I'm down

dr-frmr:
Update with @tacryt-socryp: this works perfectly and now it's just a matter of pre-hashing smart.hoon and building a cache for contracts so they only need to be hashed once. We discussed adding a gas decrement on every traversed hash, not just those that aren't in the cache. This will fix a potential infinite loop issue and incentivize sequencers to proactively cache hashed contracts, not the opposite.

tacryt-socryp:
Update: smart.hoon has been fully pre-hashed and infinite loop issue is fixed. Merging into master now.

PR #62, opened 2022-06-03
Mill: add multiple continuation calls

dr-frmr:
TODO test

### DIFF
--- a/app/sequencer.hoon
+++ b/app/sequencer.hoon
@@ -219,7 +219,7 @@
       =/  =height
         .^(height %gx /(scot %p our.bowl)/ziggurat/(scot %da now.bowl)/block-height/noun)
       =/  our-chunk=chunk
-        %+  ~(mill-all mil me (need town-id.state) height now.bowl)
+        %+  ~(mill-all mil me (need town-id.state) height)
           town.state
         ~(tap in basket.state)
       ::  currently clearing mempool with every chunk, but
--- a/app/ziggurat.hoon
+++ b/app/ziggurat.hoon
@@ -547,7 +547,7 @@
       =+  /(scot %p our.bowl)/wallet/(scot %da now.bowl)/account/(scot %ux (need address.state))/(scot %ud relay-town-id)/noun
       =+  .^(account:smart %gx -)
       =/  globe-chunk
-        (~(mill-all mil - relay-town-id height.state now.bowl) globe.state ~(tap in basket.state))
+        (~(mill-all mil - relay-town-id height.state) globe.state ~(tap in basket.state))
       =+  :_  height.state
           %+  ~(put by (~(gut by queue.state) slot-num ~))
           relay-town-id  globe-chunk
--- a/lib/zig/mill.hoon
+++ b/lib/zig/mill.hoon
@@ -144,16 +144,17 @@
       ++  fertilize
         |=  =yolk
         ^-  embryo
-        ?.  ?=(account caller.yolk)  !!
+        ::  this stops contracts from grabbing grains they hold!
+        ::  ?.  ?=(account caller.yolk)  !!
         :+  caller.yolk
           args.yolk
         %-  ~(gas by *(map id grain))
         %+  murn  ~(tap in my-grains.yolk)
         |=  =id
-        ?~  res=(~(get by granary) id)      ~
-        ?.  ?=(%& -.germ.u.res)             ~
-        ?.  =(holder.u.res id.caller.yolk)  ~
-        ?.  =(town-id.u.res town-id)        ~
+        ?~  res=(~(get by granary) id)         ~
+        ?.  ?=(%& -.germ.u.res)                ~
+        ?.  =(holder.u.res (pin caller.yolk))  ~
+        ?.  =(town-id.u.res town-id)           ~
         `[id u.res]
       ::  +germinate: take contract-owned grains in egg and populate with granary data
       ++  germinate
@@ -186,13 +187,28 @@
         ?~  gan=(harvest p.u.chick to.p.egg from.p.egg)
           [~ ~ rem %7]
         [`p.u.chick gan rem err]
-      ::  hen result, continuation
-      =*  next  next.p.u.chick
-      ::  continuation calls can alter grains
-      ?~  gan=(harvest roost.p.u.chick to.p.egg from.p.egg)
+      ::  hen result, continuation calls
+      =/  next  next.p.u.chick
+      =/  gan  (harvest roost.p.u.chick to.p.egg from.p.egg)
+      |-
+      ?~  gan
+        ::  harvest checks from last call failed
         [~ ~ rem %7]
-      %-  ~(incubate farm u.gan)
-      egg(from.p to.p.egg, to.p to.next, budget.p rem, q args.next)
+      ?~  next
+        ::  all continuations complete
+        ::
+        [`roost.p.u.chick gan rem %0]
+      ::  continue continuing
+      ::
+      =/  intermediate
+        %-  ~(incubate farm u.gan)
+        egg(from.p to.p.egg, to.p to.i.next, budget.p rem, q args.i.next)
+      ?.  =(%0 errorcode.intermediate)
+        [~ ~ rem errorcode.intermediate]
+      %=  $
+        next  t.next
+        gan   final.intermediate
+      ==
       ::
       ::  +weed: run contract formula with arguments and memory, bounded by bud
       ::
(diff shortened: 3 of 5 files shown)


dr-frmr:
NOTE: the set of owned grains stored in a `wheat` is not currently updated. Need to add part to mill that does this or simply get rid of this field

dr-frmr:
Ok, continuation calls of the singly-linked variety (just one at a time) seem to be working well, got the mill testfile working well again. Now need to test a list of continuation calls and see how granary gets updated as the list is traversed.

dr-frmr:
One thing I'm thinking about a lot as I look at mill and make changes: I think it makes sense to assign gas fees to some of our arms. This is just a note for now, since our gas fees aren't important yet and the budgeting machinery for this will be very easy to insert into arms like `germinate` and `fertilize`. But it will become important, because the computational work required to search out and attach various data `rice` to a contract call will eventually be significant and worth guarding against abuse.

tacryt-socryp:
Note that all granary modifications must occur at the end, not prior, and that scries are fixed to the beginning state. I know you're aware, just explicitly stating this constraint. It would make sense that later continuation calls may see what grains have been modified and issued already by prior calls, perhaps passed in via the cart?

dr-frmr:
> Note that all granary modifications must occur at the end, not prior, and that scries are fixed to the beginning state. I know you're aware, just explicitly stating this constraint. It would make sense that later continuation calls may see what grains have been modified and issued already by prior calls, perhaps passed in via the cart?

Right now the `+incubate` arm is what we're using in the recursive execution of continuation calls, so the granary is currently being modified between each call. I agree it has to change since it breaks our shared environment across all calls in a block. 

So yeah, maybe this is what we do to make `mem` into something useful? It could be a granary-type, named something which evokes its ephemeral nature, and just contain any interim results stored in the `roost` of a `hen`? Could be pretty cool. Then, at the end, it gets validated in a batch and unified with `granary`?

tacryt-socryp:
Got it. I'll think more on this, as we need to strike a balance between expressiveness and our ability to parallelize execution and reads.

tacryt-socryp:
Ok. More or less, we need to refactor the mill entrypoint into two steps. The first should be an explicit loop that executes every transaction and ends up with three result items for each one: the amount of gas used, the json hint file for feeding into zock, and the (already validated) diff to perform on the granary. Then the second step will be an explicit "reducer" step that takes the list of transactions, checks if any touched the same state, then throws away the least valuable ones if so, then applies the diffs to the granary and outputs the JSON hints to the file system to eventually be fed into Zock.

dr-frmr:
Ok this sounds really good and is something I can get started on pretty much right away!

tacryt-socryp:
Imo we have to execute the lower priority one in the next chunk by virtue of the fact that the semantics of the computation may be totally different once the grain in question is modified and may produce unintended effects.

dr-frmr:
Yeah, you could send tokens of different types fine, but in this execution model, you wouldn't be able to have two txns which both send `zigs` or any other token to the same account. The only way to resolve this would be switching to a UTXO model where each token send creates a new `rice`, I think.

A lot of DeFi applications rely on the ability to chain together multiple interactive txns in the same block (flash loans for example). I can see moving the token standard to a UTXO-style model for parallelization's sake, but I'm concerned that this inability for two transactions to "accumulate" information into one piece of state will cause a lot of issues across many smart contract applications we want to support...

tacryt-socryp:
That's fair, but I'm relatively confident we could play with "one to n chunk" length time-lock contracts to allow composable atomicity. There will be tradeoffs in contract complexity, but I really think we need to make the sacrifices necessary to get these parallel execution characteristics, as ETH's "serial executor" model is an easy pitfall to find ourselves in if we take a misstep.

dr-frmr:
I suppose that if there was a way to guarantee a transaction's continuation-calls would actually happen across multiple blocks, you could accept the intermediate outputs on a block-by-block basis and perform things without the usually required collateral.

tacryt-socryp:
Other tactics than timelock contracts for introducing atomicity are to allow sending to 1-of-n addresses that have the same contract code or owner, allowing frequently used contracts or frequently sent to individuals to "shard" their grains to allow a corresponding increase in simultaneous interactions per chunk.

dr-frmr:
I’ll have to think about this more and maybe write something up tomorrow or Monday — I see the benefits of full enforced parallelization but I’m also really nervous that it will introduce too much complexity in building the products with proven demand if we don’t do it right — maybe that means full UTXO-ization of all contracts, or something else.

dr-frmr:
Okay, this is the way I've laid out the problem and our two proposed solutions:

**Option 1** (presented only for comparison): Execute all transactions one-by-one in an order determined by gas price.

**Option 2:** Make use of our data model to easily parallelize transactions which don't modify the same state.
- **Option 2a:** While parallelizing, take transactions within the block which do touch the same state and order them by gas price. This will preserve the computational benefits of parallelization, but lead to similar dynamics as serial execution in ETH.

- **Option 2b:** Enforce full parallelization within a block, such that no two transactions can be included that modify the same state. Continue to use gas price as priority, but push the lower-priority transaction(s) into subsequent blocks.
When assessing these options, the question comes down to analyzing various MEV tactics and weighing their prevention against making contract writing easy. The computation benefits are less relevant because Options 2a and 2b both result in enough "batching" of work to take plenty advantage of multicore processing.

Classic MEV relies on scanning the mempool for pending transactions, finding addresses that are to receive a valuable token, swapping in your own address, and running that virtually to see if it would be profitable for you. That modified transaction is then submitted advantageously to the mempool, either with a higher gas cost than the original or (if you're actually the miner/validator for that block) just processing it first.

There are more "layered" tactics which are prevented by forcing every transaction to work against the state generated from the previous block, such as sandwiching an atomic swap. But enforcing that only one transaction per block can modify a certain piece of state won't stop these attacks. Frontrunning will still work the same way, and layered transaction tactics will only be slowed down. The fundamental problem won't change, and it's possible that this new execution method will reveal novel MEV strategies -- the people running these programs are some of the smartest developers/mathematicians in the world and I'm more comfortable with "the devil we know" in this case.

Finally we have to weigh the cost of accomodating this full parallelization. It would complicate every contract -- any piece of data that is intended to be accessible by multiple parties would have to fracture itself. For tokens, this could be accomplished via a UTXO model, but there's a whole host of data units (multisig wallets, liquidity pools, DAO member sets) for which this will require hard thinking to work around. I fear that this will create too steep a learning curve to development on Uqbar, and we'll lose out to technically simpler chains. There's an appeal to the purity and rigor of having each transaction operate "flatly" against a shared state, but I don't actually see what advantages it brings.

I think we can gain all the computational advantages of parallelization by using our existing data model to intelligently batch together transactions that don't touch overlapping state. This isn't the place to solve MEV, rather, that's one of the places where our identity/reputational infrastructure will shine. With permissionless entry to sequencing on a town, it should always be possible to find a reputable ship to receive your transactions which are vulnerable to frontrunning.

dr-frmr:
After spending a while thinking about this yesterday and today I think that with enforced parallelism we'd basically not prevent any common MEV tactics while making it way harder to replicate most of the proven-use-case products on ETH

timlucmiptev:
> After spending a while thinking about this yesterday and today I think that with enforced parallelism we'd basically not prevent any common MEV tactics while making it way harder to replicate most of the proven-use-case products on ETH

Yeah, @tacryt-socryp I follow MEV stuff extremely closely, and also know a lot of prior art on the lines of the approach you described. I'm happy to talk about it with you more later this week, but @dr-frmr is correct on this one, and this is not the vector from which to do MEV minimization or capture.

Put another way: this is a dimension that Urbit actually doesn't give us differentiation in solving, which means that the prior research, experience and thinking in this field is extremely relevant and not sidesteppable by us.

dr-frmr:
> Ok. More or less, we need to refactor the mill entrypoint into two steps. The first should be an explicit loop that executes every transaction and ends up with three result items for each one: the amount of gas used, the json hint file for feeding into zock, and the (already validated) diff to perform on the granary. Then the second step will be an explicit "reducer" step that takes the list of transactions, checks if any touched the same state, then throws away the least valuable ones if so, then applies the diffs to the granary and outputs the JSON hints to the file system to eventually be fed into Zock.

So based on our conversations, this still looks like a good plan -- we just need to take in a variable number of "passes" and send those overlapping transactions to the next pass until those have been exhausted, yes?

dr-frmr:
Gonna close this draft since the work has been picked up by me in PR #80

PR #64, opened 2022-06-06
Remove unused generators and lib files

ynx0:
Removes unused generator `blue.hoon`, as well as `bink.hoon` files.


dr-frmr (approved):

dr-frmr:
@tacryt-socryp is this cool with you, now that we've gone straight to zink?

tacryt-socryp:
Fine

PR #66, opened 2022-06-09
indexer: scries give null if not found

hosted-fornet:
Behavior change requested by @willbach . Now scrying on a valid path for an item that does not exist (for example, a grain that does not exist), will return `~` in Urbit or `null` in JSON.

Removed the `/has-*` scry paths, since a user can now just scry directly for the item, and should check if the `update:indexer` returned is `~` to determine if it exists or not.

@dr-frmr , PTAL specifically at `app/wallet.hoon`, `app/ziggurat.hoon`, and `lib/wallet/util.hoon`. I had previously broken code when switching to use the `/has-*` scry endpoints, and so am cautious now that I am changing it back.

An alternative method to this PR would be to make scries return `(unit update:indexer)` rather than add `~` as a valid `update:indexer`. I chose to do it this way because it makes the en-/de-JSON simple. We can also potentially use the `~` `update:indexer` down the line if we want to implement a heartbeat, where, e.g., the indexer sends `~` to subscribers if their subscribed object has not changed, and a non-`~` update on change.


dr-frmr (approved):
Looks like you covered all the wallet scenarios that used `has` path, nice

tacryt-socryp (approved):

PR #68, opened 2022-06-10
Mill: add salt and type checks to changed grains in +harvest

dr-frmr:
Quick change to make sure `salt`s are immutable on `rice` (@ynx0 reminded me that this behavior is important and should be enforced) and that the type of a grain is not being swapped in a change event.

### DIFF
--- a/lib/zig/mill.hoon
+++ b/lib/zig/mill.hoon
@@ -263,14 +263,20 @@
           `(~(uni by granary) (~(uni by changed.res) issued.res))
       ?&  %-  ~(all in changed.res)
           |=  [=id =grain]
-          ::  id in changed map must be equal to id in grain AND
           ::  all changed grains must already exist AND
+          ::  new grain must be same type as old grain AND
+          ::  id in changed map must be equal to id in grain AND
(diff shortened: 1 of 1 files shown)


lib/zig/mill.hoon
```diff
@@ -263,14 +263,20 @@
...
+          ::  if rice, salt must not change AND
           ::  no changed grains may also have been issued at same time AND
           ::  only grains that proclaim us lord may be changed
-          ?&  =(id id.grain)
-              (~(has by granary) id.grain)
-              !(~(has by issued.res) id.grain)
-              =(lord lord:(~(got by granary) id))
+          =/  old  (~(get by granary) id)
+          ?&  ?=(^ old)
+              ?.  ?=(%& -.germ.u.old)
```
tacryt-socryp, line 274: More simply written as `?:  ?=(%& ...)` imo. Was a little confusing to read

dr-frmr: sure, will change

dr-frmr:
@tacryt-socryp this is good to merge right -- I fixed the conditional ordering

PR #70, opened 2022-06-13
mill: comment out zink printf causing compilation fail

ynx0:
Without this, starting a testnet fails when running `|install our %zig`

cc @dr-frmr


dr-frmr:
found this too, no idea how it got in that last PR.. let's delete that whole comment block

PR #71, opened 2022-06-15
contracts: use library imports

hosted-fornet:
Make use of library imports added in PR #52 to avoid duplicating type definitions in `+write` and `+read` arms and to hold various utility functions (including `enjs` for `+read` arms).

Currently working.

One thing I'd like input on is whether we should provide some namespacing conventions within contract libraries. This is dependent also on whether the `/=` import system is the "final form" or just a stepping stone to a more complete import system. There are two things I am thinking about along these lines.

1. Nested/recursive imports and namespacing. For example, it would be nice to have both a `sur` and a `lib` file for complicated contracts. And this can work -- provided you don't need to import the `sur` file into the `lib` file. One solution to this problem is to make the import parser recurse through imported files. I think there are some subtleties here that may be hard to get right. I understand also that flat imports forces devs to save space on chain, and maybe this is a good thing. A hack we can use right now to get nice namespacing is to have a convention that flat import `lib` files are divided internally into cores. To make this concrete, imagine a `lib` file for a hypothetical contract divided into type and code cores:

```
|%  
++  sur
  |%
  +$  foo  @p
  +$  bar  @ud
  --
++  lib
  |%
  ++  baz
  ^-  bar:sur
  2
  --
--
```

This gives us the option for nice namespacing we'd get from the full Hoon import system.

2. Importing stdlib stuff. We still cannot import stdlib -- and maybe this is by design. The current workaround is to copy-paste the stdlib code into your contract `lib` (within a core?). I do think this would be nice to have and is probably not to crazy to imagine people wanting to make use of stdlib molds and such within contracts.

Thoughts @dr-frmr @ynx0 ? If dividing contract `lib`s into cores sounds reasonable, I can add commits to this PR to suggest/demo a convention.


tacryt-socryp:
What stdlib stuff do you want that you don't have?

hosted-fornet:
In the dao case, `resource`. Easy enough to copy-paste into the contract. Maybe copy-pasting is the general solution: it is nice in that it restricts you from doing things that would bloat your contract (like multiply-importing the same code, e.g., the types into the contract and the contract lib); it is not nice in that copy-pasting code into a contract lib seems to violate some conventional wisdom of DRY.

tacryt-socryp:
Well, resource is not in the stdlib. It's in a particular sur file, sur/resource.

hosted-fornet:
Ah, pardon my incorrect terminology. In the case of `resource` it is pretty trivial to copy-paste, so perhaps this is a problem to address another day, if it really is a problem. I guess it just goes against my intuition to copy-paste code as opposed to importing it, but there are some legitimate advantages to be had here for copy-pasting.

dr-frmr:
If people want to use a sur/lib/etc structure I think this is a good way to enable it, but I hesitate to make it the encourages default way of writing contracts — I think many applications will only want/need one file. And once we get interest from people wanting to spin up a special purpose town to run an MMO or something on chain, we can fill out the build system to be fully featured. This is  proof of concept more or less to guarantee that will be possible in the future.

dr-frmr:
> I guess it just goes against my intuition to copy-paste code as opposed to importing it, but there are some legitimate advantages to be had here for copy-pasting.

To me, it's an open question whether sharing code between on- and off-chain applications is good or bad. In a traditional blockchain environment, using anything not purpose-built and optimized for that specific gas-bounded execution environment would be wildly inefficient compared to the alternative. But it's possible that our sharded+ZK environment is truly so lush with compute resources that the developer time and ease becomes the resource bottleneck. So I want to wait and see what becomes the bottleneck before committing to either eschewing DRY principles in favor of traditional smart contract dev strategies or fully replicating ford for contracts.

hosted-fornet:
c8c6ed6 shows what I mean by splitting contract `lib`s into cores.

ynx0:
My 2c: sur and lib combined are not so bad, especially when you don't have a crazy amount of types / arms (which we want to encourage anyway). `smart.hoon` is maybe a good example of this. Separate cores in one library file are a cool in-between if the need truly arises, but I'd echo @dr-frmr here and also might say that it's probably overkill as a default pattern.

Another thought: when would you want to import only the sur but not the lib (or even vice versa)? The only main example I can think of is if you actually have multiple "lib" files that all import the sur but not each other. This seems like a valid use case but for something much further down the line. Thus we probably don't want to encourage it right now and instead want to exert downward pressure on contract/lib/sur size, even if it's somewhat arbitrarily.

dr-frmr:
@hosted-fornet I'm down with you merging this btw, looks good as is -- though I reserve the right to mess with the zigs contract further, lol. dachus-tiprel is also gonna take a hand to the token/NFT contracts once he gets up to speed.

PR #72, opened 2022-06-15
Zane (name TBD): unified read/write "vane" for dApps

dr-frmr:


dr-frmr:
@hosted-fornet I've stubbed out the zane vane here. I want to provide subscribes that transparently forward data from indexer(s), scries for one-time data reads (which can subscribe to the indexer path, get that first update served, then close the subscription), and an interface for raw contract writes (eggs) and reads directly from a sequencer.

hosted-fornet:
On it

dr-frmr:
Sweet this is coming together fast

hosted-fornet:
Okay, as of d3f0aad have: subscribes that forward data from indexer; scries for one-time data reads from indexer. I can continue to work on this, but those are the parts that touch the indexer based on your comment above. LMK

tacryt-socryp:
Can we just call it "uqbar.hoon"? That seems to be the most direct way to say "Uqbar API"

dr-frmr:
> Can we just call it "uqbar.hoon"? That seems to be the most direct way to say "Uqbar API"

Yeah, I'm down for that, or any other name suggestion -- could we discuss with everyone in dev-main and see what people think there

dr-frmr:
Just want to clarify that I switched town IDs to `@ux` to bring them in line with everything else. This was foolish of me since it would have been better as its own quick ctrl+F PR.

hosted-fornet:
> Just want to clarify that I switched town IDs to `@ux` to bring them in line with everything else. This was foolish of me since it would have been better as its own quick ctrl+F PR.

This is a good change. One thing I noticed, though, is that there is a `town` type, and that many of the other types have a `town=id:smart` attribute. I'd prefer if the `id`-typed attributes be called something different from town to avoid confusion, say `town-id` like when they were still `@ud`s.

dr-frmr:
Ok, we can call it town-id or just id when it refers to that, yeah.

dr-frmr:
I broke something in this branch or in PR 69 and now contracts won't run through mill -- looking into why, might have to roll back some stuff to a working version..

lib/zig/sys/smart.hoon
```diff
@@ -107,7 +107,7 @@
       status=@ud  ::  error code
```
hosted-fornet, line 107: `town-id.shell` should be updated to `@ux`

dr-frmr: yup

dr-frmr:
ugh, tried changing commit message but just added a new one. `@ud` -> `@ux`

dr-frmr:
Oh also changed `cart` to have `now` instead of `block`

PR #77, opened 2022-06-20
Add instructions on how to recompile contracts and the stdlib

ynx0:
I can also remove https://github.com/uqbar-dao/ziggurat/blob/d395f3bb8100ddbfad10c38cd8e7606545e164d3/lib/zig/sys/create-smart-library.md if it makes sense.


dr-frmr:
Don't think there's any need to have this whole blob copy-pasted in readme. Let's just have the 1-line smart-lib compile and a single example of using the deploy generator. Can replace the DAO contract compile guide with this too.

ynx0:
@dr-frmr PTAL

ynx0:
No, I did a dumb and committed to the wrong branch. Lemme fix.

dr-frmr (approved):

PR #78, opened 2022-06-21
wallet: fix `egg-status-code` type by using `errorcode` directly

ynx0:
Haven't tested the actual code but I tested equivalent code in the dojo and it should work.


dr-frmr (approved):
I just added `%8` so will need to resolve a little conflict, but approved

PR #83, opened 2022-06-27
Adding test arms for fungible.hoon

witfyl-ravped:
Adding a few test arms for %deploy, %set-allowance, and %mint

### DIFF
--- a/tests/contracts/fungible.hoon
+++ b/tests/contracts/fungible.hoon
@@ -91,7 +91,7 @@
   =/  =embryo
     :+  owner-1
       `[%set-allowance 0xcafe 10]
-      (malt ~[[id:account-1 account-1]])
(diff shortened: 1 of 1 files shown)


tests/contracts/fungible.hoon
```diff
@@ -141,18 +186,156 @@
...
+
+  (expect-eq !>(res) !>(correct))  
+
+
+::  tests for %deploy
 ::
+++  test-deploy  ^-  tang
+  =/  token-salt
+    (sham (cat 3 0xbeef 'TC'))
+
```
dr-frmr, line 288: Hoon should never have empty lines -- if you need a visual separator, use a line with only `::`
Everything other than that looks great.

dr-frmr (changes requested):

dr-frmr (approved):

PR #84, opened 2022-06-27
Add ship wl

marcoleder:
dr/rollup-mode is missing util.hoon in lib/zig, hence forked master -> adjusted wl -> PR into dr/rollup-mode


dr-frmr:
The whitelist is now located here in the rollup branch: https://github.com/uqbar-dao/ziggurat/blob/4f146ff264a6076df049ad29810770eccda21c79/lib/rollup.hoon#L11

Can you modify that one instead? Don't want to bring the old method in, lot of conflicts here.

Also in future we usually name branches with this format: `[initials]/[name]` to keep things organized

PR #69, opened 2022-06-13
Uqbar Rollup Testnet: %sequencer rewrite, remove %ziggurat, create %uqbar: read/write "zane"

dr-frmr:
This PR removes the consensus engine that was `%ziggurat`. It replaces consensus with a rollup contract simulator, which is the `%rollup` agent. 

We've also added `%uqbar`, an agent that acts as a vane for Uqbar interactions. 

Finally, there are a number of changes to the execution engine that have made their way into this PR. `mill.hoon` now emits a proper state transition, including hints for Zink. Town IDs now match the rest of the chain in being `@ux`. Contracts have a changed context, where `now` has replaced `blocknum`.

Mill still needs to be parallelized, which will entail restructuring the `mill-all` arm to be a big `turn` over the list of txns, and at that point we can get the `diff` granary we need to provide to the rollup.

##  Uqbar.hoon

```
      (a)       (b)            (c)
Users <--> Zane --> Sequencers --> Rollup
             ^      |  (e)        |
         (d) |      V             | (f)
             Indexers   <----------
```
Basic APIs / interfaces

(a) Users <--> Zane
- Gall dApps use zane to query their contract address, perform reads, and submit transactions (signed via wallet)
- Zane allows apps to track selected grains and subscribe to get those changes

(b) Zane --> Sequencers
- Zane submits transactions directly to sequencer (it gets the information to do so from Indexer)

(c) Sequencers --> Rollup
- Sequencers post batches to rollup
- Batches contain everything needed for ZK-verifiable state transitions

(d) Indexers --> Zane
- Zane chooses a ship hosting an indexer to serve as a data source
- Indexer provides requested information along with mapping of town to sequencer (so zane knows who to submit to)

(e) Sequencers --> Indexers
- Indexers subscribe to one or more sequencers to get latest state (note that batches have replaced blocks, but same concept)
- Indexers verify received state against state roots posted to rollup

(f) Rollup --> Indexers
- Indexers subscribe to the rollup contract (currently simulated by a gall agent) to get latest state roots for desired towns, so they can verify state received by sequencers

---

##  Uqbar Rollup Testnet Spec

Goal: take out all consensus machinery from ziggurat and sequencer apps, and replace them with a unified rollup sequencer designed off this spec: web+urbitgraph://group/~datwet/uqteam/graph/~datwet/dev-posts-685/170141184505683749127804303000964431872
(spec pasted here for posterity, can edit as implemented):

[...]

PR #85, opened 2022-06-27
Contract type upgrades

dr-frmr:


lib/zig/contracts/zigs-parallel.hoon
```diff
@@ -0,0 +1,106 @@
...
+++  write
+  |=  inp=embryo
+  ^-  chick
+  ?~  action.inp  !!
+  =/  act  ;;(action u.action.inp)
+  ?-    -.act
+      %give
+    =/  giv=grain  +.-:grains.inp
+    ?>  &(=(lord.giv me.cart) ?=(%& -.germ.giv))
+    =/  giver=account  ;;(account data.p.germ.giv)
```
dr-frmr, line 22: Cannot for the life of me figure out a good way to generalize this pattern into an arm for smart.hoon.

Requirements are that it should take a `grain`, assert that it's a `rice`, and mold it to a type. The catch is that it has to preserve the `rice` status of that grain so you can use salt/data after calling, and generally be easier to understand for a noob than what we have here now.

hosted-fornet:
Will take a closer look at this once I've finshed with PR #69. The first thing that jumps out at me, though, is that changes to the `zigs.hoon` contract and lib will have to be merged with master at some point, so might be good to take a look at the state on master so that merge is as easy as possible.

dr-frmr:
> the `zigs.hoon` contract and lib will have to be merged with master at some point, so might be good to take a look at the state on master so that merge is as easy as possible.

I have this PR high up the "food chain" -- will first merge in rollup PR, then mill parallelization, then this.

Unfortunately it contains some breaking changes for all contracts so we'll have to deal with that then

ynx0:
Changes look pretty good, and I like the `call`/`continuation`/`result` arms at first glance. Will have to try them out.


Since this is a QoL PR anyways, I'd like to suggest one more change:  I think would be great for mill to automatically construct the `(map id grain)`s for the contract writer and make e.g. `rooster`s become `[changed=(list grain) issued=(list grain) =crow]` (can still keep harvest id assertions). This avoids them the hassle of calling `malt`/`gas:by` every time.


This arm should do the trick, name notwithstanding:

```hoon
++  grainz
  |=  grz=(list grain)
  %-  ~(gas by *(map id grain))
  %+  turn  grz
  |=  [=grain]
  [id.grain grain]
```

As an intermediate solution to allow a gradual refactor, we could simply add this arm to `smart.hoon` 
without doing the type change immediately, and then have contracts slowly update their code. 

Then in the second phase we do the type change and hopefully the refactor is a bit easier. This approach has has it's own pros and cons, I guess.

lib/indexer.hoon
```diff
@@ -124,10 +124,8 @@
   ++  yolk
     |=  =yolk:smart
     ^-  json
-    ?>  ?=(account:smart caller.yolk)
     %-  pairs
-    :~  [%caller (account caller.yolk)]
-        [%args ~]
+    :~  [%args ~]
```
hosted-fornet, line 128:
```
:~ [%args ~]
```
 ->
 ```
:~ [%action ~]
```

File: lib/zig/contracts/zigs-parallel.hoon [hunk 3]
```diff
@@ -0,0 +1,145 @@
...
+    ::  all other grains in grains.inp are tickets
+    =^  spent=(list grain)  rec
+      %^    spin
+          ~(val by (~(del by grains.inp) my-account.act))
+        rec
+      |=  [=grain =account]
+      ^-  [^grain ^account]
+      ?>  &(=(lord.grain me.cart) ?=(%& -.germ.grain))
+      =/  tik  ;;(ticket data.p.germ.grain)
+      ?<  redeemed.tik
```
hosted-fornet, line 73: Should this instead allow redemption of the other tickets? E.g. something like

```
?:  redeemed.tik
  [grain account]
```

dr-frmr: Yeah, that makes a lot of sense.

lib/indexer.hoon
dr-frmr: Will change

lib/zig/contracts/zigs.hoon
```diff
@@ -13,98 +14,99 @@
...
-    ?+    args  !!
+    ?+    path  !!
         [%rice-data ~]
       ?>  =(1 ~(wyt by owns.cart))
       =/  g=grain  -:~(val by owns.cart)
       ?>  ?=(%& -.germ.g)
-      ?.  ?=([@ @ @ @ ?(~ [~ @]) ? ?(~ ^) @ @] data.p.germ.g)
+      ?:  ?=(%account label.p.germ.g)
         (account:enjs:lib ;;(account:sur data.p.germ.g))
       (token-metadata:enjs:lib ;;(token-metadata:sur data.p.germ.g))
```
dr-frmr, line 99: much cleaner

ynx0: very nice. kind of synonymous with how gall agents tag vases with `@tas` "marks", so it's nice to see some precedent there

lib/zig/contracts/lib/fungible.hoon
```diff
@@ -1,38 +1,50 @@
...
-        salt=@            ::  data added to hash for rice IDs of this token
-                          ::  (currently hashed: symbol+deployer)
+        salt=@            ::  salt value for rice holding accounts of this token
     ==
   ::
   +$  account
     $:  balance=@ud                     ::  the amount of tokens someone has
         allowances=(map sender=id @ud)  ::  a map of pubkeys they've permitted to spend their tokens and how much
         metadata=id                     ::  address of the rice holding this token's metadata
+        nonce=@ud                       ::  necessary for gasless approves
```
ynx0, line 27: When I was first introduced to the gasless signing code, I remember being confused by the collision between nonce as used in fungible and the normal per-pubkey nonce. 

It seems like *nonce* is definitely the correct term here so maybe renaming it to `fungible-nonce` might help (open to better suggestions).

Don't feel too strongly about it but thought it was worth mentioning

lib/zig/mill.hoon
```diff
@@ -120,89 +203,104 @@
...
+      ::  +fertilize: take yolk (contract arguments) and populate with granary data,
+      ::  and iff the action is to the zigs contract, insert budget.
       ++  fertilize
-        |=  =yolk
+        |=  [from=id =yolk]
         ^-  embryo
-        ?.  ?=(account caller.yolk)  !!
-        :+  caller.yolk
-          args.yolk
+        :-  ?~  action.yolk  ~
```
ynx0, line 290: I see that in `embryo` the action comes unwrapped which is nice. But what does it mean for a user to construct a yolk with a null action?

lib/indexer-bowl.hoon
```diff
@@ -140,19 +140,17 @@
...
-    =/  args=json
-      ?~  args.yolk  ~
+    =/  action=json
+      ?~  action.yolk  ~
       %:  scry-contract-read-arm
           our.bowl
           to.shell
           %egg-args
-          (scot %ud (jam u.args.yolk))
+          (scot %ud (jam u.action.yolk))
```
tacryt-socryp, line 149: If this is only used as a part of a scry path, I recommend rendering it as `@ux` using the `en:base16` arm because it's jetted and will take up less space.

lib/zig/sys/smart.hoon
```diff
@@ -184,4 +225,3192 @@
...
+    ::                                                  ::  ++as-octt:mimes:html
+    ++  as-octt                                         ::  tape to octstream
+      |=  tep=tape  ^-  octs
+      (as-octs (rap 3 tep))
+    --
+  --  ::  html
+::
+::  CRYPTO from zuse.hoon
+::
+++  crypto-non-zuse  ^?
```
tacryt-socryp, line 543: Why not just crypto?

tacryt-socryp (approved):
Let's discuss the cores you've included (and their structure) tomorrow. Ideally I'd like to separate things out into cores for better organization while keeping everything in this file. Otherwise LGTM

ynx0:
You may or may not be aware of this already, but I'm getting a bunch of these when I try to run `-test ~[/=zig=/lib/mill]`

```
fund: in lsh, parent 18226962 not found at 7
fund: in end, parent 18226962 not found at 7
fund: in bex, parent 18226962 not found at 7
```

which iirc are jet mismatches.

tacryt-socryp:
`+mule` will not be able to find jets because our standard library is a different shape than the normal one.

ynx0:
Ah I see. So, not a problem.

lib/zig/mill.hoon
dr-frmr: It means they're calling a contract that doesn't care what the user inputs, and will only ever do one thing. Can potentially remove unit from yolk in the future, but that would be a larger refactor

lib/zig/sys/smart.hoon
dr-frmr: Ok, works as intended with just `crypto`

lib/zig/contracts/lib/fungible.hoon
tadad: yeah this isn't a bad idea. dr and I also talked about making it into a `(map id @ud)` which takes away the ordering constraint. E.g. let's say you want to let 15 people take 10 zigs from you. Right now this doesn't work unless all 15 take their tokens in the same order that you signed them in. But a mapping would let people take whenever they like, regardless or ordering

PR #94, opened 2022-07-08
fungible: fix logic in `%mint` ensuring `mintable` is correctly set

ynx0:
I think new behavior is correct but would like a double check on it. I should note that I haven't done a thorough pass over the contract for correctness, but this still stood out to me.

### DIFF
--- a/lib/zig/contracts/fungible.hoon
+++ b/lib/zig/contracts/fungible.hoon
@@ -144,7 +144,7 @@
       =.  data.p.germ.tok
         %=  meta
           supply    new-total
-          mintable  ?:(=(u.cap.meta supply.meta) %.y %.n)
+          mintable  (lth supply.meta u.cap.meta)
         ==
       ::  for accounts which we know rice of, find in owns.cart
       ::  and alter. for others, generate id and add to c-call


dr-frmr:
Looks good -- note that these lines do have the same effective logic since we're already asserting that cap is gth new-total. In fact, do we need to use `new-total` here since we're updating `supply` in the same `=:`?

What we really need is tests for the mint action. `tests/contracts/fungible.hoon` has a lot missing.

ynx0:
Yeah, good point.  Also I just realized I accidently made the change to fungible.hoon instead of nft.hoon because they're so similar (at least in this regard). I'll probably write tests for `nft.hoon` at some point if no one else gets to it first.

One additional thing I noticed in nft.hoon was that in
https://github.com/uqbar-dao/ziggurat/blob/73e263890f93c13cd00a148c88dfe7666c086893/lib/zig/contracts/nft.hoon#L98-L108

there is no check to make sure that the caller of `%mint` is actually in the `minters` field of the metadata. We should probably do that.

dr-frmr:
> I'll probably write tests for nft.hoon at some point if no one else gets to it first.

If you're gonna spend time on patching untested contracts, that time would be just as well applied generating test coverage, lol. The reason it hasn't been done yet is because the contracts are proof of concept drafts, and so their interfaces are subject to many changes. They're not written to be production ready. Productionizing them will absolutely require 100% test coverage just like any good smart contract.

ynx0:
Understood. I was looking through the nft contract anyways to slim it down and had just noticed this, and, thinking there was just a small change to be made I thought might as well. But it’s now that it’s clear that there’s more to be done aka proper test coverage I think a new future PR is probably the way to go.

PR #100, opened 2022-07-19
Dr/iota typing

dr-frmr:
Introduces `iota` and `lump` (complex iota) types to smart.hoon. These types allow contracts to publish their `interface` and `types` on-chain, which contain maps of their action types and rice types, respectively. These types are represented by `lump` because molds cannot be stored in any space-reasonable way. This restricted type definition should be effective for all contract contexts -- it covers every possible inert data representation, the restrictions are to remove things like lazy execution and core structures.

**As a result, contract `read` arms no longer need to be used for typing rice and arguments.** These arms can still potentially be useful for people seeking to read directly from chain, but now applications can simply look at the `lump` for that rice `label` or action head and generate either a type or json pattern with which to represent that data. Note that this automatic conversion has not yet been implemented and is in fact a bit tricky for things like list/set/map but we'll figure it out later.

Scrys inside contracts have been built out here -- in theory, everything is connected, but final implementation is dependent on storing state as merkle tree and setting this up will happen in next PR.

Removed some of `crypto` core that wasn't relevant. Let's see if we can remove even more -- what are the specific crypto functions we need available inside contracts?

**Reviewers**: main focus here is changes to `smart.hoon`, `zink.hoon` -- other files' changes are mostly to align apps with new data structures.

## Breaking changes to contracts!

- rather than taking in an `embryo`, the `write` arm now takes in a custom-defined `action` type union. Branch off the head `@tas` of this type like in an on-poke arm.
- the entire `yolk`/`embryo` structure in `egg`s has been reworked: `yolk` is now `(pair @tas *)`, which is interpreted as an action by a contract. The `egg` signature has also been taken out of `shell` and now sits in the new `egg` def: `[=sig =shell =yolk]`. this is because the signature has to hash and sign all the data in both shell and yolk.
- `wheat` now store two maps of `@tas` to `lump` -- make sure to generate these when you publish a contract or spawn one in an init generator.
- `grain` has been reworked to hopefully provide less of a hassle when unwrapping `rice` for use in contracts. It is now directly `(each rice wheat)`, and `rice` and `wheat` both separately contain the metadata properties such as `id` and `lord`.
- the old pattern of extracting a data from a rice and then `;;`ing it is deprecated in favor of using `+husk`, which also checks for lord/holder ownership if desired
- the `now` field in `cart` which contained a `@da` has been replaced by the provable `batch` number.

This is the last chance for breaking changes to `smart`. After this PR we intend to freeze `smart` until public testnet launch, with the exception of converting certain maps to merkle maps.


ynx0:
The one thing that jumps out to me, especially considering the imminent freezing of smart (and likely cooling of other parts of the system) is that arguments in `yolk` are still `ream`ed by mill without any `map` or `set` constructor arms available.

Creating these without `malt`/`silt` or `gas:by`/`gas:in` seem like they would be error prone and poor UX.

I think now that the types are quite restricted, it makes sense to include some facility (i.e. bringing them into the subject thru smart.hoon) to easily construct maps and sets, which would otherwise have to be done manually through nock-js or something.

dr-frmr:
> arguments in yolk are still reamed by mill without any map or set constructor arms available.

It's not in `mill` that this issue occurs but rather with the wallet app, right? Once it's inside Hoon the action is just a noun and can be constructed with those constructors.

I thought we solved the wallet issue by doing this: `q:(slap !>(+:(cue q.q.smart-lib)) (ream yolk.act))`
I agree that's not *great*, but it doesn't actually touch the execution engine so we can continue to iterate on that.

ynx0:
Understood. I want to run through a use case for this so bear with me here:

Let's say I want to create a multisig with an initial set of members. I would like to do something like:
```hoon
[%submit-custom from to town gas=[1 10.000] action='[%create-multisig thresh=3 members=(silt ~[~zod ~nec ~bus])]'
```

(or maybe `members=[[%set %p] (silt ~[~zod ~nec ~bus])]`? just riffing here)




But unless `silt` is part of smart-lib it looks like that's not really possible. Ditto for `malt`. 
Maybe I'm missing something though.

dr-frmr:
@ynx0, `malt`, `silt`, and `gas` are all contained in hoon.hoon which is built into smart-lib -- you can do exactly what you have there in your action.

lib/zig/contracts/zigs.hoon
```diff
@@ -12,77 +12,66 @@
...
+    ?~  to-account.act
+      ::  if no receiver account specified, generate new account for them
+      =/  =id  (fry-rice me.cart to.act town-id.cart salt.giver)
+      =/  =rice
+        [salt.giver %account [0 ~ metadata.data.giver] id me.cart to.act town-id.cart]
+      =/  next  [%take to.act amount.act [%grain id.giver] `[%grain id.rice]]
+      (continuation [me.cart town-id.cart next]^~ (result ~ [%& rice]^~ ~ ~))
+    ::  have a specified receiver account, grab it and add to balance
+    =+  (~(got by grains.cart) id.u.to-account.act)
+    =/  receiver  (assert-rice account:sur - `me.cart `to.act)
```
ynx0, line 57: Bikeshed time!

`assert-rice` is a decent name for this arm but I feel like there could maybe be a better one, since you're not just doing asserts on the rice but also typecasting it and producing the updated rice as well.

The best one I could think of was actually a metaphor: `(husk-rice ...)` or even simply `husk`. It seems evocative enough of what's going on but is also not too precise enough to be disqualified.

(Dis)honorable mentions:
- `pull-rice`
- `get-rice`
- `validate-rice`
- `cast-rice`
- `typed-rice`
- `checked-rice`

Feel free to leave as is though. It's obviously not a dealbreaker.

dr-frmr: I like `husk` a lot too, will change, thanks

app/sequencer.hoon
```diff
@@ -139,20 +139,19 @@
...
-      ^+  basket
-      %-  ~(run in eggs.act)
-      |=  =egg:smart
-      [`@ux`(sham q.egg) egg]
+      :_  state(basket (~(uni in basket) received))
+      %+  turn  ~(tap in received)
+      |=  [hash=@ux =egg:smart]
+      ^-  card
+      =/  usig  (ecdsa-raw-sign:secp256k1:secp:crypto `@uvI`hash (need private-key.state))
+      =+  [%uqbar-write !>([%receipt hash (sign:sig our.bowl now.bowl hash) usig])]
```
tacryt-socryp, line 153: Always cast vases. :)

app/uqbar.hoon
```diff
@@ -195,7 +195,7 @@
     =/  path  ~[/track/[i.t.wire]]
     :_  this
     ?~  p.sign  ~
-    [%give %fact path %write-result !>([%rejected src.bowl])]~  
+    [%give %fact path %write-result !>([%rejected src.bowl])]~
```
tacryt-socryp, line 198: Always cast vases.

lib/indexer-bowl.hoon
```diff
@@ -138,21 +138,11 @@
     ==
   ::
   ++  yolk
-    |=  [=shell:smart =yolk:smart]
+    |=  [=yolk:smart]
```
tacryt-socryp, line 141: brackets unnecessary with one arg

lib/merk.hoon
```diff
@@ -9,8 +9,8 @@
 ++  shag                                                ::  256bit noun hash
   |=  yux=*  ^-  hash  ^-  @
   ?@  yux
-    (shas %gash yux)
-  (shas %shag (jam yux))
+    (pedersen-hash %gash yux)
```
tacryt-socryp, line 12: won't compile

lib/zig/contracts/zigs.hoon
```diff
@@ -12,77 +12,66 @@
...
+    =:  balance.data.giver     (sub balance.data.giver amount.act)
+        balance.data.receiver  (add balance.data.receiver amount.act)
+        allowances.data.giver  %+  ~(jab by allowances.data.giver)
```
tacryt-socryp, line 60: use queen-side for this jab: https://developers.urbit.org/reference/hoon/style

lib/zig/mill.hoon
```diff
@@ -273,61 +275,26 @@
...
-        ?.  =(lord.u.res find)          ~
-        ?.  =(town-id.u.res town-id)    ~
-        `[id u.res]
-      --
+        ?:  ?=(@ux from.shell.egg)  [from.shell.egg 0]
+        [id.from.shell.egg nonce.from.shell.egg]
+      ::  insert budget argument if egg is %give-ing zigs
+      =?  q.yolk.egg  &(=(to.shell.egg zigs-wheat-id) =(p.yolk.egg %give))
+        [budget.shell.egg q.yolk.egg]
+      ?~  gra=(~(get by granary) to.shell.egg)  [~ ~ ~ ~ budget.shell.egg %5]
```
tacryt-socryp, line 284: put the result on next line

lib/zig/mill.hoon
```diff
@@ -273,61 +275,26 @@
...
+      ?.  ?=(%| -.u.gra)                    [~ ~ ~ ~ budget.shell.egg %5]
```
tacryt-socryp, line 285: same

lib/zig/mill.hoon
```diff
@@ -369,24 +341,35 @@
...
+        ::  note that we can actually push whatever vased noun into the contract
+        ::
         ?:  test-mode
           ::  run without zebra
+          ~&  >  "yolk: {<yolk.egg>}"
+          ~&  >>>  "cart: {<cart>}"
           =/  res
-            :-  (mule |.(;;(chick q:(shut dor %write !>(cart) !>(embryo)))))
-            (sub budget (mul rate.p.egg 7))
+            :-  (mule |.(;;(chick q:(shut dor %write !>(cart) !>(yolk.egg)))))
```
tacryt-socryp, line 362: Let's switch the testmode flow to use `+mock` and `+ajar` instead of `+mule` and `+shut`. That way we can input our scry function `+search` into both of them so that scry shares semantics across test and real modes!

lib/zink/pedersen.hoon
```diff
@@ -64,6 +64,7 @@
     |=  [a=@ b=@]
     ^-  @
     |^
+    ~&  >>>  "hoon..."
```
tacryt-socryp, line 67: remove

tacryt-socryp (changes requested):
Structure good, have a few nits and a one actual ask regarding `+mule` flow.

dr-frmr:
@tacryt-socryp: converted all maps that operate inside mill to merkle. Then went through and converted uses of those merkle trees throughout the codebase. Ran some tests through mill and it all works! I think it's time to get a fresh hash-cache.noun so we can run zebra at speed.

dr-frmr:
Swapped in `husk` for `assert-rice` @ynx0

ynx0:
> wheat now store two maps of @tas to lump -- make sure to generate these when you publish a contract or spawn one in an init generator.

Question about this: how would a dev do this in practice? Is it just manually writing out the types or importing them and including them in the wheat? Reason I ask is because it's mentioned but I don't see any tooling to support it and it's unclear exactly what is required.

Ambivalent on whether tooling is necessary before merge, since it seems the important thing is to just freeze smart.hoon.

dr-frmr:
Jake is working on tooling for lumps. Yeah, we don't actually need the auto-conversion methods right now, but will get them in the next several weeks for sure.

app/sequencer.hoon
```diff
@@ -174,11 +173,12 @@
       ::  publish full diff data
       ::
       ::  1. produce diff and new state with mill
+      =/  batch-num  1  ::  TODO
```
ynx0, line 176: Looks like this todo needs to be done in the current PR.

dr-frmr: true -- added

ynx0 (approved):

dr-frmr:
Switching to draft til I get a scry working

lib/zig/sys/smart.hoon
```diff
@@ -22,6 +22,12 @@
       ==
   p.grain(data ;;(typ data.p.grain))
 ::
+::  +scry: scry wrapper
+::
+++  scry
+  |=  =id
+  .^(grain /granary/id)
```
ynx0, line 29: This should be `(scot %ux id)` but more importantly I don't think this wrapper arm is necessary yet. `.^` are fine, even from a stylistic point of view imo. 

It seems a bit early to make this wrapper.

tacryt-socryp: I desire the wrapper arm as a means of further documenting the semantics of how we do scries within contracts, and also as an alternative to `.^`.

lib/zig/sys/smart.hoon
```diff
@@ -25,71 +51,131 @@
...
-++  zigs-wheat-id  `@ux`'zigs-contract'  ::  hardcoded "native" token contract
 ::
-+$  account    [=id nonce=@ud zigs=id]
-+$  caller     $@(id account)
-::  currently only using ecdsa [v r s]
-::  +$  signature  [r=@ux s=@ux type=?(%schnorr %ecdsa)]
+++  zigs-wheat-id  `@ux`'zigs-contract'   ::  hardcoded "native" token contract
+::
++$  account  [=id nonce=@ud zigs=id]
++$  caller   $@(id account)
```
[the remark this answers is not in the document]
dr-frmr: We could get rid of the fork entirely. It's sort of a leftover option for when we were considering how to have contracts make calls themselves. It's pretty clear now that we only use the `account` structure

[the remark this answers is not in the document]
ynx0: That's what I was hoping for. Sweet.

dr-frmr:
Scrys are solid. Had to remove `mock` mode from mill due to it not handling scrys correctly (perhaps an issue to bring up sometime).

lib/zig/sys/smart.hoon
```diff
@@ -18,78 +46,129 @@
...
-    (~(gas by *(map id grain)) (turn burned |=(=grain [id.grain grain])))
+  :^    (gas:(bi id grain) *(merk id grain) (turn changed |=(=grain [id.p.grain grain])))
+      (gas:(bi id grain) *(merk id grain) (turn issued |=(=grain [id.p.grain grain])))
+    (gas:(bi id grain) *(merk id grain) (turn burned |=(=grain [id.p.grain grain])))
   crow
 ::
+::  +continuation: generate a chick containing an intermediate result and a list of next calls
+::
 ++  continuation
   |=  [next=(list [to=id town-id=id args=yolk]) rooster=chick]
```
ynx0, line 63:
```suggestion
++  continuation
  |=  [next=(list [to=id town-id=id =yolk]) rooster=chick]
```

Just to be consistent

lib/zig/sys/smart.hoon
```diff
@@ -1,10 +1,38 @@
...
+::  return the rice with the data inside asserted into the mold given.
 ::
-++  fry-contract
+++  husk
+  |*  [typ=mold =grain lord=(unit address) holder=(unit address)]
+  ?>  ?&  ?~(lord %.y =(lord.p.grain u.lord))
+          ?~(holder %.y =(holder.p.grain u.holder))
+          ?=(%& -.grain)
+      ==
+  p.grain(data ;;(typ data.p.grain))
```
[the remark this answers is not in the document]
dr-frmr: We should return all the data, otherwise you have to be repetitive to access certain info. If I can do the first option and it returns a grain that doesn't have to be branched on again to assert it's a rice, I'll switch to that.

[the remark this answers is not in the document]
dr-frmr: Just tried returning a typed grain from `husk`, and it works, but I don't really feel like it's better because you're stuck with the `p` branching -- I kind of prefer getting a rice and just attaching the `each` tag when I want to return it. Since `husk` is only for rice and should crash on wheat I think this is reasonable. Dealing with a grain out of `husk` actually adds a lot of extra characters to a contract..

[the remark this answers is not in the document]
ynx0: Gotcha

app/indexer.hoon
```diff
@@ -640,7 +640,6 @@
       =:  egg-index         (gas-ja-egg egg-index egg town-id)
           from-index        (gas-ja-second-order from-index from town-id)
           grain-index       (gas-ja-batch grain-index grain town-id)
-          grain-eggs-index  (gas-ja-second-order grain-eggs-index grain-eggs town-id)
```
hosted-fornet, line 643: I'm curious: why was grain-eggs was removed?

dr-frmr: Well, eggs no longer contain any grains -- we've changed the old `embryo` structure and replaced all the contract input grain maps with scrying the granary directly.

hosted-fornet: Oh, I see. It is probably desirable to have grain-eggs indexed at some point, if possible, but I'm not familiar enough with the zink code and scrying the granary to see how to do it. I'll put it on my todo list for a future PR.

ynx0:
One issue that cropped up for me that I thought I should mention to potentially save you the goose chase.

Here is the buggly line: `(husk account:sur giv me.cart id.from.act)`.

Looks fine, but actually it should actually be ``(husk account:sur giv `me.cart `id.from.act)``.

This is because of the way that wet gates work I think. What happens is the error shows up only at the definition of `husk` in smart.hoon and makes it look like it's erroneously defined, with no mention of the faulty callsite anywhere. Kinda cringe.

lib/zig/sys/smart.hoon
```diff
@@ -1,10 +1,38 @@
+::  ::                                         ::::::    ::::::    ::    ::::::
+::  ::  smart.hoon: contract standard library  ::  ::    ::  ::  ::  ::  ::  ::
+::  ::                                         ::  ::    ::::    ::::::  ::::
+::  ::                                         ::  ::    ::  ::  ::  ::  ::  ::
+::  ::  five: contract functions               ::  ::    ::  ::  ::  ::  ::  ::
+::::::                                         ::::  ::  ::::::  ::  ::  ::  ::
+=<
 |%
 ::
```
ynx0, line 9: Petition to add `$big` to smart.hoon somewhere. It's needed in places where you are explicitly not using `+result` and are trying to do e.g. `(put:big issued last-grain)` or `(gas:big ...)` etc.
```suggestion
::
++  big  (bi id grain)  ::  merkle engine for granary
```

app/indexer.hoon
dr-frmr: I think the best way might be to lightly interpret the action contained in the egg and turn any grain-ids inside to links to that grain, could even do purely on frontend potentially.

dr-frmr:
@ynx0 yes, `husk` takes units for its final two arguments.

lib/zig/sys/smart.hoon
dr-frmr: @tacryt-socryp does this seem right for pre-including a merkle map engine, or would users be better off making it themselves?

tacryt-socryp: Go for it, no harm, and it gives it a canonical name.

lib/zink/zink.hoon
```diff
@@ -317,37 +309,42 @@
   ++  jet
     |=  [tag=@ sam=*]
     ^-  book
-    ?.  (~(has in jets) tag)
+    ?:  ?=(%slog tag)
+      ::  ignore trace printfs
```
tacryt-socryp, line 313: I was wondering what we'd do about this lol

lib/zig/sys/smart.hoon
```diff
@@ -18,78 +46,129 @@
...
+      [%sb @sb]  [%sc @sc]  [%sd @sd]  [%si @si]
+      [%sx @sx]  [%sv @sv]  [%sw @sw]
+      [%da @da]  [%dr @dr]
+      [%f ?]     [%n ~]
+      [%if @if]  [%is @is]
+      [%t @t]    [%ta @ta]  ::  @tas
+      [%p @p]    [%q @q]
+      [%rs @rs]  [%rd @rd]  [%rh @rh]  [%rq @rq]
+      ::  contract types
+      [%address @ux]  [%grain @ux]
```
ynx0, line 143: Now that `plant` is gone and grains are scried out from the granary, is the `%grain` case/distinction necessary here?

Similar question wrt `%address`.

dr-frmr: They will remain iota types but the [%grain-id] tag that we have in contract actions will go away now that it's not being interpolated.

ynx0: I see. Still wondering why these are necessary instead of just using `%ux`. Like are `%grain` iota validated in some way by checking if they are in the granary or not before executing the contract? Are `%address` iota checked to be valid addresses (or intended to be)?

 A quick search doesn't yield me any usages.

dr-frmr:
Alright, unfortunately the `husk` wet gate is not functional inside zink. Something about the mold structure being passed around.

ynx0:
Sad!

lib/zig/sys/smart.hoon
dr-frmr: they're not necessary they're just nice to have because these iotas are used to specify contract arguments and it's good to show what that argument is being used for. same reason we alias `@ux` to `id`, `address`, and `hash` in various places.

tacryt-socryp (approved):
Beautiful. LGTM

lib/zink/zink.hoon
```diff
@@ -395,10 +395,17 @@
...
-
+        %turn
+      ::  TODO: determine how best to validate complex jet inputs
+      ::  this will crash if the input is bad.
+      %&^(some (slum turn sam))
     ::                                                                       ::
     ::  etc                                                                  ::
     ::                                                                       ::
+        %need
+      ?.  ?=((unit) sam)  %|^trace
```
tacryt-socryp, line 490: Not enough... this can still crash if the `unit` is `~`.

dr-frmr: fixed

PR #103, opened 2022-07-25
Refactor indexer+uqbar+wallet to match new types

dr-frmr:
- refactor uqbar.hoon
- get rid of germ type and some egg stuff in indexer, replace with new types JSON representation
- add interface/action type retrieval scries to sequencer


app/sequencer.hoon
```diff
@@ -288,18 +286,50 @@
...
+    =/  id     (slav %ux i.t.t.path)
+    =/  label  (slav %tas i.t.t.t.path)
+    ?~  town  [~ ~]
+    ?~  g=(get:big p.land.u.town id)
+      ::  contract not found in state
+      ``noun+!>(~)
+    ?.  ?=(%| -.u.g)
+      ::  found ID isn't a contract
+      ``noun+!>(~)
+    ?~  typ=(~(get by interface.p.u.g) label)
```
ynx0, line 322: Looks like this should be `types.p.u.g`

lib/indexer.hoon
```diff
@@ -121,22 +121,21 @@
     ==
   ::
   ++  yolk
-    |=  =yolk:smart
+    |=  [=yolk:smart]
```
ynx0, line 124: Technically wrapping this expr in `[]` is unnecessary, even though tbh I think stylistically I prefer it.

ynx0 (changes requested):

lib/indexer.hoon
tacryt-socryp: I stylistically do not prefer it.

dr-frmr: I don't like this either and will remove -- this ends up happening often for me cause I'll refactor an arm to only have 1 input and forget to remove the brackets, and since it doesn't introduce a compile error...

dr-frmr:
Addressed the review, think we're good

ynx0 (approved):

PR #106, opened 2022-07-30
sequencer: update `:sequencer|init` to work with new types

hosted-fornet:
Was trying to play with zink/zock today and so updated `gen/sequencer/init.hoon` to work with new types. I have elided `interface` and `types` from `wheat`s.


dr-frmr:
Thank you -- gotta go through all this stuff and update. Unfortunately haven't hooked up enough jets in zink / worked out the jet hinting stuff quite enough to get zigs contract working yet, but very close.

dr-frmr (approved):

PR #107, opened 2022-07-31
merk: add `+key` to `bi`

ynx0:
Ended up being useful during debugging. Adapted from `key:by` so it should be correct.

### DIFF
--- a/lib/merk.hoon
+++ b/lib/merk.hoon
@@ -257,5 +257,12 @@
       $(a r.a, p.q.n.b (mer b [p q.q]:n.b))
     =.  r.b  $(b r.b, l.a ~, p.q.n.a (mer a(l ~) [p q.q]:n.a))
     $(a l.a, p.q.n.b (mer b [p q.q]:n.b))
+  ++  kee
+    |=  [a=mert]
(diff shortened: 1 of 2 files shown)


dr-frmr:
Nice, can we call it `key` though? Just to keep in line with other operations having same names as `map`.

ynx0:
Oh yeah should’ve said but if you do that everything breaks because it
clashes with the `key` mold defined earlier.

dr-frmr:
Ah, I see. I'm inclined to say we change the internal type in order to keep the tree interface the same..

ynx0:
Ok so I just realized that apparently there are two different instances of `merk` and `bi`, one in `merk.hoon` and one in `smart.hoon`, and that sent me on a goose chase for a bit.

Which one do we want to keep? I guess both is also an answer in which case I could modify both for the time being but that seems undesirable long term.

dr-frmr:
Yes, both are necessary. The version inside the smart contract library must never be used outside it, since the jet tagging conventions are different and you will end up running merk operations completely unjetted, which takes forever.

dr-frmr (approved):
This looks good

PR #108, opened 2022-08-03
Dr/wallet redux

dr-frmr:
Updates:

(wallet)
- wired requests through uqbar.hoon (still relies on having indexer on same ship due to no remote scry)
- major refactor of utils/parsing
- refactor of token-store, metadata-store
- self-held token/account/metadata types set to match new zigs.hoon, nft.hoon contracts
- fixed bugs in status codes updating

(non-wallet)
- light refactor to conq
- publish contract tests written
- nft contract rewritten, one nft per rice now :)
- added info to readme, generated real addresses with seeds for test usage

This is a more polished wallet app. After merge, will go ahead and match up with fabnev for a couple days to re-align frontend.


lib/wallet/util.hoon
```diff
@@ -53,69 +53,78 @@
...
+  %=  $
     grains-list  t.grains-list
+    new-book  (~(put by new-book) [town-id.p.grain id.p.grain] asset)
   ==
 ::
-::  TODO: replace this whole janky system with a contract read that returns the type of the its rice
+++  discover-asset-mold
+  |=  data=*
+  ^-  asset
+  =+  tok=(mule |.(;;(token-account data)))
```
tacryt-socryp, line 76: Note that this is directly equivalent to running `+soft`, and the code would be neater that way.

dr-frmr: oh dang okay, nice

dr-frmr:
@hosted-fornet do you want to do code review here or should I just merge and have you fix anything bad? Whatever is easier for you

hosted-fornet:
Go ahead and merge

PR #110, opened 2022-08-08
Update publish.hoon and its tests

dr-frmr:
Nothing special here, just updated the code in our publish contract and got the tests running through mill+zink. Need more tests, obviously, can add those at a later date.


dr-frmr:
Closing cause I brought this into my wallet refactor branch

PR #111, opened 2022-08-08
fungible: rework and make tests pass

ynx0:
This PR refactors fungible.hoon to work with mill as of #100. 

Remaining items:
- [ ] get all old tests working
- [ ] (potentially) factor out fungible logic to be shared between fungible/zigs/nft


dr-frmr:
With regards to factoring out shared logic, most sensible way atm I think would be to put a couple token functions in a library and import it in multiple contracts. We should set up a way to publish *just* a library on-chain but not ready for that yet. Post-testnet.

PR #112, opened 2022-08-10
readme: updates and cleanup

hosted-fornet:


hosted-fornet:
Two TODOs near the end. I'll think about what else needs to be on here.

Another thing to think about: what are we calling this repo, now that the dev environment is `%ziggurat`? Some prose in the intro will need to be changed -- for now it is just referred to as ziggurat.

dr-frmr (approved):

PR #113, opened 2022-08-10
Add maps and sets to smart that use Pedersen

dr-frmr:
Realized when setting up our new and improved NFT contract the other day that we didn't have a way to use sets and maps in contracts, since they use `mug` for item ordering, which is not an effective hashing function to prove. Therefore, we need to create alternative versions that use Pedersen or another proving-friendly hash.

I added two types to smart.hoon, since I didn't want to just overwrite the existing set and map definitions. We may want to remove those from our hoon.hoon though eventually. The new types are `pset` and `pmap`. They have engines `pn` and `py` respectively. I've made use of these in the NFT contract and they seem to work just fine. Of course, we'll want to jet `shag`, and probably `sore` and `sure` too (these are used in psets and pmaps to perform ordering).

- rebuilt smart-lib.noun
- rebuilt hash-cache.noun to match
- ran mill-2 tests, passing
- made test for NFT, passing


tacryt-socryp:
Honestly, would be pretty cool here to just remove `+map` and `+set` and their operations from `hoon.hoon` and move them to `smart.hoon` if that's feasible without messing up later layers of `hoon.hoon`. The names `map` and `set` are semantically pleasing and people are used to using them, and there are no semantic differences (the hash function changing is an implementation detail).

hosted-fornet (approved):
LGTM, runs as expected.

dr-frmr:
> Honestly, would be pretty cool here to just remove `+map` and `+set` and their operations from `hoon.hoon` and move them to `smart.hoon` if that's feasible without messing up later layers of `hoon.hoon`. The names `map` and `set` are semantically pleasing and people are used to using them, and there are no semantic differences (the hash function changing is an implementation detail).

@tacryt-socryp, this is what I wanted to do but then I got nervous about people being able to use these maps and sets on-and-off-chain. Since the hashing functions would be different, what would happen if I pulled the data of a grain into my gall app and tried to put/get from the map?

tacryt-socryp:
That's fair. There would (there will regardless) need to be a conversion function, and I can imagine people getting confused of converting between two things of the same name. Valid to keep them both!

tacryt-socryp (approved):
Consider adding in `+jar` and `+jug` equivalents prior to mainnet release, even if they are not yet being used. LGTM

dr-frmr:
I'll add pedersen equivalents for `jar`, `jug`, `silt`, `malt`, and others in a fresh PR, will merge this

PR #115, opened 2022-08-15
Dr/fungible

dr-frmr:
Building off of @ynx0's branch here: #111 

- finish off fungible contract
- align fungible tests
- fix supply-cap bug in nft contract
- switch zigs contract to using pmaps (important!)
- fix bug in publish contract
- delete leftover old tests for multisig/dao contracts

NOTE: the ridiculous diff is due to deleting DAO test file + reducing size of compiled fungible by 57%
DAO tests are worth saving, but can remove from test suite in main ziggurat repo. TODO make separate repo for DAO project.

getting closer to the finished contract suite.. more tests!


ynx0:
One refactor I'd done to nft.hoon when writing the nft-slim contract was that I'd gotten rid of `mintable` on token-metadata and made `cap` a regular `@ud`, replacing that property with a `/mintable` query.

Thought it might be worth mentioning as a potential refactor since I always thought the current way is a bit error prone and feels off.

dr-frmr:
> One refactor I'd done to nft.hoon when writing the nft-slim contract was that I'd gotten rid of `mintable` on token-metadata and made `cap` a regular `@ud`, replacing that property with a `/mintable` query.
> 
> Thought it might be worth mentioning as a potential refactor since I always thought the current way is a bit error prone and feels off.

I'm not a fan of requiring a scry to determine a property that we can keep easily in metadata -- I like how I have it now, but did have to make a change because the way I originally was managing `cap` was wrong.

PR #116, opened 2022-08-15
Add asset escape hatch to `mill`

dr-frmr:
Finally, burns will be real. With this addition to `mill`, one can submit a special `%burn` transaction to a sequencer.

The burner must be either `holder` or `lord` of the grain in question.

A burn transaction is marked by setting the `to` contract ID to `0x0` and the action to `[%burn {id} {town-id}]`, where {id} is of the grain in question, and {town-id} is the town the burned grain is destined for.

Burns done this way produce an event noting the burn.

What else needs to be added here? So far, we're just creating the burn -- the "reconstitute" transaction will come later, as we're waiting to see how to gather that information from the rollup. Burns will be one-way until then!

### DIFF
--- a/lib/zig/mill.hoon
+++ b/lib/zig/mill.hoon
@@ -266,7 +266,7 @@
       ^-  hatchling
       =/  from  [id.from.shell.egg nonce.from.shell.egg]
       ::  check for grain burn transaction
-      ?:  &(=(to.shell.egg 0x0) =(p.yolk.egg %burn))
+      ?:  &(=(0x0 to.shell.egg) =(%burn p.yolk.egg))
         (cook egg)
       ::  insert budget argument if egg is %give-ing zigs
       =?  q.yolk.egg  &(=(to.shell.egg zigs-wheat-id:smart) =(p.yolk.egg %give))
--- a/lib/zig/mill.hoon
+++ b/lib/zig/mill.hoon
@@ -267,7 +267,7 @@
       =/  from  [id.from.shell.egg nonce.from.shell.egg]
       ::  check for grain burn transaction
       ?:  &(=(0x0 to.shell.egg) =(%burn p.yolk.egg))
-        (cook egg)
+        (poach egg)
       ::  insert budget argument if egg is %give-ing zigs
       =?  q.yolk.egg  &(=(to.shell.egg zigs-wheat-id:smart) =(p.yolk.egg %give))
         [budget.shell.egg q.yolk.egg]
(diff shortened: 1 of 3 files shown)


lib/zig/mill.hoon
```diff
@@ -273,6 +265,9 @@
       |=  [=egg:smart hits=(list hints) burned=^granary]
       ^-  hatchling
       =/  from  [id.from.shell.egg nonce.from.shell.egg]
+      ::  check for grain burn transaction
+      ?:  &(=(to.shell.egg 0x0) =(p.yolk.egg %burn))
```
hosted-fornet, line 269: nit: consider rewriting this line to

```
      ?:  &(=(0x0 to.shell.egg) ?=(%burn p.yolk.egg))
```

lib/zig/mill.hoon
```diff
@@ -433,6 +430,37 @@
...
+    ::
+    ++  cook
+      |=  =egg:smart
+      ^-  hatchling
+      ::  TODO provide new error codes for these things
+      ::  TODO assign reasonable fixed cost for a burn
+      =/  fixed-burn-cost  1.000
+      ::  charge fixed cost for failed transactions too
+      ::  TODO should do this everywhere that we can inside +farm
+      =/  fail  [~ ~ ~ ~ (sub budget.shell.egg fixed-burn-cost) %6]
```
hosted-fornet, line 445: I like this idiom.

lib/zig/mill.hoon
dr-frmr: True. Always forget that ordering

lib/zig/mill.hoon
dr-frmr: Ultimately we’ll be able to dynamically calculate all these costs, as we will have to expand the part of mill that `zebra` wraps around

ryjm: `cook` is already in `hoon.hoon` and used pretty frequently - maybe `boil` or `bake`? `grill`?

lib/zig/mill.hoon
```diff
@@ -433,6 +430,37 @@
               !=(zigs.from id)
           ==
       ==
+    ::
+    ::  +cook: handle special burn-only transactions, used for manually
+    ::  escaping some grain from a town. must be EITHER holder or lord to burn.
+    ::
+    ++  cook
```
ryjm, line 437: `cook` is already in hoon.hoon and used pretty frequently - maybe `boil` or `bake`? `grilll`?

dr-frmr: well, you don't bake an egg. poach it is

dr-frmr:
Want to merge this now and add town-nonces later, if that sounds good to @hocwyn and @workinglikeadogalldayonuqbar

PR #119, opened 2022-08-17
Hf/uqbar wallet polish 0

hosted-fornet:
220821

Main changes here were to %uqbar. These include:

1. Generalizing some of the scry and subscription paths,
2. Add logic to periodically check whether indexer sources are live, and to subscribe to most-likely-to-be-live indexers

Additional things that it would be nice to do with the fallback logic:
1. Move existing subscriptions from unresponsive indexers to live indexers. This could be done today.
2. Fallback in case of failed scries to most-likely-to-be-live indexers. The logic for fallback is mostly here, but this cannot be done yet since we don't have remote scry -- currently the only indexer that will work for scries is on the same ship as %uqbar.

My plan is to point a new PR at this one that will route chain write requests through %uqbar to %wallet, since this is a desired feature with more time pressure. After that I'll take another look at applying fallback to subscriptions.


hosted-fornet:
Shoot, I broke something here as of a2fdfe4. Will fix and push a commit asap.

hosted-fornet:
Fixed in ec53d53

lib/indexer.hoon
```diff
@@ -3,6 +3,20 @@
 /+  smart=zig-sys-smart
 ::
 |%
+::  use, e.g., to extract single grain from a
+::  `/newest/grain/...`  scry
+++  extract-single-grain
```
dr-frmr, line 8: Is it not possible to set up the `newest` path to just return `(unit grain)` rather than the `update:ui` type? That way we could get rid of this stuff and just parse the single grain

dr-frmr: @hosted-fornet other than this q, approved btw

hosted-fornet: I like the unified `update:ui` return type -- aside from just being predictable and easy for a dev to remember, it makes passing things through %uqbar a lot easier and nicer. How about a compromise: add add'l sub-types to `update` like

```
[%newest-grain =grain]
[%newest-egg =egg]
...
```

Then the logic to get the grain is something like
```
?~  scry-result=.^(update:ui foo)  !!
?.  ?=(%newest-grain -.scry-result)  !!
(bar grain.scry-result)
```

as opposed to
```
?~  scry-result=.^(update:ui foo)  !!
(bar u.scry-result)
```

hosted-fornet: Accidentally added `%newest-*` on PR #125 60d5816 rather than here

dr-frmr (approved):

PR #120, opened 2022-08-17
multisig.hoon: update

dr-frmr:
Clean up the multisig contract and write thorough tests for it. Still not 100% coverage here, but pretty good.

No longer a generic contract -- people must deploy the contract for their multisig, reason being that the contract needs to custody funds.


dr-frmr:
merging

PR #123, opened 2022-08-18
fix hint order

workinglikeadogalldayonuqbar:


lib/zink/zink.hoon
```diff
@@ -537,28 +536,31 @@
   ::
   ++  hash
     |=  n=*
-    ^-  [(unit phash) appendix]
+    |^  ^-  [(unit phash) appendix]
```
tacryt-socryp, line 539: You don't need this `|^` as you neither include more arms nor have a second recursion point.

PR #127, opened 2022-08-31
give zigs account rice nonces

dr-frmr:
draft -- we want zigs.hoon to match the fungible standard, so i'm adding nonces to the rice


dr-frmr:
will merge after %ziggurat

PR #130, opened 2022-09-06
wallet: make transactions go through signing step

dr-frmr:
This PR makes the wallet transaction flow correct. Rather than %submit and %submit-custom pokes going straight through the wallet, we change to using a single %transaction poke, which creates an unsigned `egg` and allows the wallet frontend to approve this transaction.

This makes match the behavior of transactions using a hardware wallet and those using the hot wallet feature inside %wallet.

The readme has been changed to match the new pokes, and show the JSON for wallet.


hosted-fornet (approved):
Code LGTM, have not tested :+1:

dr-frmr:
I tested everything in the readme and works great, so gonna merge

PR #131, opened 2022-09-06
sequencer: add modules to do automatic batching

dr-frmr:
Goal here is to make a couple pre-made apps that perform various sequencing strategies. The two I plan to make are: batching regularly on an interval, and batching after the sequencer's "mempool" has a certain number of transactions in it. We can later add more advanced batching strategies once we have towns set up to run on rings of sequencers.


dr-frmr:
@hosted-fornet these look good to you? Got one that triggers every interval, and one that regularly polls sequencer for mempool size, triggering at certain threshold.

hosted-fornet (approved):
These seem like they should work. Two main comments that you can take or leave:

1. There is no way to turn these off. One possible method would be to make `interval` and `threshold` into `unit`s: if `~`, then the agent stops polling until they are changed to non-`~`.
2. It might be better if these were combined into one agent with multiple modes. I don't think this is particularly urgent. But if we don't go that route yet, it might be nice to give them more descriptive names, e.g., `batcher-inverval` and `batcher-threshold` or something rather than `1` and `2`.

dr-frmr:
Oh, I should definitely add a way to turn them off other than just nuking the agent. And I’ll rename them to those more descriptive names

dr-frmr:
I thought about just doing one agent but I figure these will actually get pretty advanced in the future so best to leave as modular as possible

hosted-fornet:
> I thought about just doing one agent but I figure these will actually get pretty advanced in the future so best to leave as modular as possible

Makes sense :+1:

PR #141, opened 2022-09-09
Slightly better compiler errors

dr-frmr:
Change conq.hoon and lib/ziggurat.hoon to make contract compile errors that match enough, such that ziggurat frontend can parse line numbers and red-squiggly-underline the site of error.

### DIFF
--- a/lib/ziggurat.hoon
+++ b/lib/ziggurat.hoon
@@ -92,16 +92,20 @@
   |^
   =/  first  (mule |.((parse-main main.proj)))
   ?:  ?=(%| -.first)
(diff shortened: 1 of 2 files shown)


lib/ziggurat.hoon
```diff
@@ -122,7 +126,7 @@
     =/  libraries=hoon  [%clsg braw]
     :-  q:(~(mint ut p.smart-lib) %noun libraries)
     (slap smart-lib libraries)
-  ::
+  ::1
```
hosted-fornet, line 129: typo

hosted-fornet (approved):
:+1:

PR #149, opened 2022-09-13
Dr/fix text to noun with zebra

dr-frmr:
In a future PR, I will do jet reform in the version of zink we're using for the exec-only testnet.


hosted-fornet (approved):
Comments:

1. Is there any case where `50.000` gas will not be enough? Should this be user configurable?
2. We `(need p.p.res)`. Is there any case where this will be a `unit` that we should gracefully handle?

dr-frmr:
If `p.p.res` is ~, "compiling" the noun has failed -- ideally we surface this on frontend, but a dojo error works for now

50.000 gas is pretty low, so we might want to make it configurable, but it would be *such* a confusing option to surface on frontend. I'm basically using it as a time cutoff, but I'm gonna increase to `200.000`, which takes at least 15 seconds to spend and I've never seen that much spend on a contract call.

PR #151, opened 2022-09-14
Fix nft contract deploy bug

dr-frmr:


dr-frmr:
feeling foolish for this one

PR #153, opened 2022-09-15
add crypto jets to zink

dr-frmr:


dr-frmr:
Need to write a contract to test these before merging

dr-frmr:
Luckily none of our existing contracts changed with the smart-lib.noun changes here.

tadad (approved):

PR #154, opened 2022-09-15
Mill: sort by nonce in addition to gasrate

dr-frmr:


dr-frmr:
This may or may not be optimally efficient but it appears to be working properly from my tests. Needed to make sure that multiple transactions from same address are sorted correctly.

hosted-fornet (approved):

PR #155, opened 2022-09-15
fix wallet unsub

dr-frmr:
uqbar.hoon needs to unsubscribe from indexer when it gets unsubbed from by an app, otherwise you don't get that crucial first update


dr-frmr:
@hosted-fornet when the wallet sends a %leave on an indexer watch path thru `uqbar`, how can we make sure that `uqbar` leaves that path as well?

dr-frmr:
Tested and fully operational, merging

PR #165, opened 2022-09-16
reintroduce ID subs (still need these!)

dr-frmr:


dr-frmr:
Shouldn't have taken these out, only the /holder subs. Tested and reintroducing!

PR #166, opened 2022-09-17
indexer: dont dump history on sub by default

hosted-fornet:
On master, subscribing to a path dumps history of that item, unless you append `/no-init`, e.g. `/grain/0xdead.beef` gives the history of the grain as an immediate reply, while `/grain/0xdead.beef/no-init` does not.

This PR swaps the default behavior. Now `/grain/0xdead.beef` does NOT give any immediate reply, while `/grain/0xdead.beef/history` does give the history of the grain.

FYI @willbach , I figure this will break the frontend, so let me know when you are ready for this to be merged.


lib/wallet/util.hoon
```diff
@@ -32,8 +32,9 @@
   ^-  (list card)
   %+  turn  ~(tap in pubkeys)
   |=  k=@ux
-  =-  [%pass - %agent [our %uqbar] %watch -]
-  /indexer/wallet/id/0x0/(scot %ux k)  ::  TODO: remove hardcode; replace %wallet with [dap.bowl]?
+  =/  w=wire  /id/0x0/(scot %ux k)
+  =-  [%pass w %agent [our %uqbar] %watch -]
+  ;:(weld /indexer/wallet w /history)  ::  TODO: remove hardcode; replace %wallet with [dap.bowl]?
```
dr-frmr, line 37: would prefer to just do this here:
```
=-  [%pass w %agent [our %uqbar] %watch -]
/indexer/wallet/id/0x0/(scot %ux k)/history
```

dr-frmr (approved):

lib/wallet/util.hoon
hosted-fornet: I tend to prefer the programmatic form since it is one less hardcoding, and easier to maintain as a result, but happy to make the change. Changed in 005c23b.

dr-frmr: Oh, because the /id/0x0/(scot.. part is sort of custom for this specific poke while the outer structure is constant?

I see -- if the line separation could be done without the weld calls I would agree, I just have a hard time justifying to myself spending the logic on what's purely syntactical

hosted-fornet: Maybe necroposting, but:

The way I think about it, code should be written with first priority readability and maintainability. Then only if needed do you optimize and care about what the computer is doing. Most of the time IDC if computer has to do more work cause CPU cycles are cheap and human brain cycles are extremely expensive. This is part of the reason I favor namespacing (no CPU cost, big human time savings when you have to read a piece of code you are unfamiliar with), meaningful variable names (I will frequently alias with `=*` even when not needed because I think it makes it clearer for the reader), and meaningful function names. If you have an application that needs optimization, profile your readable code, then optimize the parts that are biggest slowdowns first, and proceed down the list there until it is fast enough.

Re this case, you can argue above is not readable, but I do think it is more maintainable, since it has less hardcoding.

PR #173, opened 2022-09-23
swap shag to sham

dr-frmr:
Using SHA hashing instead of the very slow Pedersen hash in order to make execution-only testnet operable in a performant way with no extra jets.


hosted-fornet (approved):
tested and lgtm :rocket:

PR #181, opened 2022-09-29
indexer: fix downtime catchup

hosted-fornet:


dr-frmr:
wow, our very first versioned state. here's to many more. tested and working, LGTM

dr-frmr (approved):

PR #186, opened 2022-10-02
Add exponential backoff to %faucet rate-limiting

dr-frmr:
This PR changes %faucet state, so we'll have to |nuke %faucet and re-instantiate. This is okay for mass-distribution, since %faucet is not in the `desk.bill` of our distribution.

### DIFF
--- a/app/faucet.hoon
+++ b/app/faucet.hoon
@@ -66,16 +66,22 @@
     ?-    -.action
         %open
       =,  bowl
+      ?:  =(%pawn (clan:title src))
+        ~|("%faucet: comets cannot use the faucet! get a planet!" !!)
+      =/  par
+        ?:  =(%earl (clan:title src))
+          (sein:title our now src)
+        src
       ?~  town-info=(~(get by town-infos) town-id.action)
         ~|("%faucet: invalid town. Valid towns: {<~(key by town-infos)>}" !!)
       =/  [unlock=@da count=@ud]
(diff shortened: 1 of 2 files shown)


dr-frmr:
@hosted-fornet let's get this in soon so certain people stop just grabbing one zig per hour. what do you think?

app/faucet.hoon
```diff
@@ -65,17 +65,26 @@
...
+        (~(gut by on-timeout) par [*@da 0])
+      ?:  (gth unlock now)
+        ~|("%faucet: must wait until after {<unlock>} to acquire more zigs." !!)
+      =/  until=@da  (add now.bowl (mul timeout-duration (pow 2 count)))
+      :_  %=    this
+              on-timeout
+            %+  ~(put by on-timeout)  par
+            [until ?:((gte count 12) count +(count))]
+          ==
+      :~  =-  [%pass /transaction-poke %agent [our.bowl %wallet] %poke -]
```
hosted-fornet, line 87: Would recommend either using `:+` here or adding a spacing `::` below to separate out the cards more clearly.

hosted-fornet (approved):
Code LGTM. I'm not entirely convinced exponential backoff is necessary; I guess the most compelling reason to do it IMO is to reduce unnecessary chain bloat. Ultimately, the code is not much more complex, and so I would say :+1:

PR #187, opened 2022-10-03
.

dr-frmr:


hosted-fornet:
fyi @dr-frmr @willbach This will break frontend if frontend relies on old subscription behavior.

willbach:
> fyi @dr-frmr @willbach This will break frontend if frontend relies on old subscription behavior.

roger. I'll update the indexer today or tmrw

PR #189, opened 2022-10-05
Da/push pull

tadad:


lib/zig/contracts/lib/fungible.hoon
```diff
@@ -175,10 +183,44 @@
...
+    =+  (need (scry u.to-account.act))
+    ::  assert that account is held by the address we're sending to
+    =/  receiver  (husk account:sur - `me.cart `to.act)
+    ::  assert that token accounts are of the same token
+    ::  (since this contract can deploy and thus manage multiple tokens)
+    ?>  =(metadata.data.receiver metadata.data.giver)
+    =.  balance.data.receiver  (add balance.data.receiver amount.act)
+    ::  return the result: two changed grains
+    %+  continuation
+      [to.act town-id.cart [%on-push id.from.cart amount.act]]~
```
tadad, line 217: `to.act` has to be a contract that implements `[%on-push from=id amount=@ud]`

lib/zig/contracts/lib/fungible.hoon
```diff
@@ -50,7 +50,7 @@
     $:  balance=@ud                    ::  the amount of tokens someone has
         allowances=(pmap address @ud)  ::  a map of pubkeys they've permitted to spend their tokens and how much
         metadata=id                    ::  address of the rice holding this token's metadata
-        nonce=@ud                      ::  necessary for gasless approves
+        nonces=(pmap address @ud)      ::  necessary for gasless approves
```
tadad, line 53: this lets you do async `%pulls`. e.g. if you are doing payroll, you can sign N messages, and have all employees `%pull` them whenever

tadad: if you have just a single @ud as a nonce, then nonce=0 needs to take, then 1, then 2, etc. which really defeats the use case.

lib/zig/contracts/lib/fungible.hoon
```diff
@@ -87,14 +88,22 @@
...
-  +$  take-with-sig
-    $:  %take-with-sig
+  +$  push
+    $:  %push
+        to=address
+        from-account=id
+        to-account=(unit id)
+        amount=@ud
+        data=*
+    ==
```
tadad, line 98: this is just an [ERC677](https://github.com/ethereum/EIPs/issues/677) implementation. very under-utilized EIP - ChainLink is the only thing I know that uses it off the top of my head.

dr-frmr: can we rename `data` to `calldata`, so as not to confuse it with `data.rice`?

tadad: it let's you send tokens *and* call a function in the same transaction. Normally have to do an `%approve` and then call the function you actually want, but with this, you can just call it once. I think most contracts *should* implement this as opposed to the `%approve` and `%take` pattern - e.g. an AMM will just have to put the relevant swap parameters in `data=*`. Of course, you can also implement this as a signature and a `%pull` (what uniswap does currently, mostly because ERC677 is not widely implemented on ETH) but this is much cleaner from both a ux and dev-ex perspective

tadad: @dr-frmr yes, done

lib/zig/contracts/lib/fungible.hoon
```diff
@@ -175,10 +184,44 @@
...
+    =+  (need (scry u.to-account.act))
+    ::  assert that account is held by the address we're sending to
+    =/  receiver  (husk account:sur - `me.cart `to.act)
+    ::  assert that token accounts are of the same token
+    ::  (since this contract can deploy and thus manage multiple tokens)
+    ?>  =(metadata.data.receiver metadata.data.giver)
+    =.  balance.data.receiver  (add balance.data.receiver amount.act)
+    ::  return the result: two changed grains
+    %+  continuation
+      [to.act town-id.cart [%on-push id.from.cart amount.act calldata.act]]~
```
tadad, line 218: `to.act` has to be a contract that implements `[%on-push from=id amount=@ud data=*]`

lib/zig/contracts/lib/fungible.hoon
```diff
@@ -196,16 +239,22 @@
       (ecdsa-raw-recover:secp256k1:secp:crypto signed-hash sig.act)
     ::  assert the signature is valid
     ?>  =(recovered-address holder.p.giv)
+    :: assert nonce is valid
+    =+  (~(get by nonces.giver) to.act)
+    ?>  .=  nonce.act
+      ?~  -
+        =>((~(put py nonces.giver) to.act 1) 0)
+      =>((~(jab py nonces.giver) to.act succ) ->)
```
dr-frmr, line 247: This could be written more concisely with `+gut`:

```
?>  .=  nonce.act
    +((~(gut by nonces.giver) to.act 0))
```

lib/zig/contracts/lib/fungible.hoon
```diff
@@ -196,16 +239,22 @@
...
     :: TODO need to figure out how to implement the deadline since now.cart no longer exists
     ?>  (lte batch.cart deadline.act)
     ?>  (gte balance.giver amount.act)
     ?~  to-account.act
     ::  create new rice for reciever and add it to state
       =+  (fry-rice to.act me.cart town-id.cart salt.p.giv)
       =/  new=grain
-        [%& salt.p.giv %account [amount.act ~ metadata.giver 0] - me.cart to.act town-id.cart]
+        [%& salt.p.giv %account [amount.act ~ metadata.giver ~] - me.cart to.act town-id.cart]
```
dr-frmr, line 255: Make sure you switch the old `@ud` nonce representation to `~` *everywhere* in the codebase. It's in the %give and %take arms, and it's also in the `zigs.hoon` contract, plus %wallet. plus `mill.hoon`. We need to match the fungible standard everywhere.

lib/zig/contracts/lib/fungible.hoon
```diff
@@ -217,8 +266,8 @@
     =:  data.p.rec  receiver(balance (add balance.receiver amount.act))
         data.p.giv
       %=  giver
-        balance  (sub balance.giver amount.act)
-        nonce  .+(nonce.giver)
+        balance   (sub balance.giver amount.act)
+        nonces    (~(jab py nonces.giver) to.act succ)
```
dr-frmr, line 270: This is equivalent to doing `(~(put py nonces.giver) to.act nonce.act)` but uses more gas

dr-frmr (changes requested):

lib/zig/contracts/lib/fungible.hoon
tadad: you do need to increment it so I think it would be `(~(put py nonces.giver) to.act +(nonce.act))`

lib/zig/contracts/lib/fungible.hoon
tadad: Fixed in `fungible.hoon` and `zigs.hoon` - but I think the nonces in `%mill` and `%wallet` are different - those are nonces related to like "how many transactions have I already sent" not "how many `%pull`s have I let people perform"

tadad:
great all the issues should be resolved

lib/zig/contracts/lib/fungible.hoon
dr-frmr: Isn't `nonce.act` asserted to be equal to whatever nonce they *had* plus one, already? Therefore if you want the nonce incrementing by one each time, you'd just insert `nonce.act`

lib/zig/contracts/lib/fungible.hoon
dr-frmr: I'm not talking about account nonces -- I'm talking about the data types that represent standard fungible tokens in wallet and mill

dr-frmr: https://github.com/uqbar-dao/ziggurat/blob/91d42d811bcca86c2acf3e0e9a963b92f557bc3a/lib/zig/mill.hoon#L204-L209

dr-frmr: https://github.com/uqbar-dao/ziggurat/blob/91d42d811bcca86c2acf3e0e9a963b92f557bc3a/sur/wallet.hoon#L123-L128

lib/zig/contracts/lib/fungible.hoon
```diff
@@ -175,53 +184,80 @@
...
+    =+  (need (scry u.to-account.act))
+    ::  assert that account is held by the address we're sending to
+    =/  receiver  (husk account:sur - `me.cart `to.act)
+    ::  assert that token accounts are of the same token
+    ::  (since this contract can deploy and thus manage multiple tokens)
+    ?>  =(metadata.data.receiver metadata.data.giver)
+    =.  balance.data.receiver  (add balance.data.receiver amount.act)
+    ::  return the result: two changed grains
+    %+  continuation
+      [to.act town-id.cart [%on-push id.from.cart amount.act calldata.act]]~
```
dr-frmr, line 218: I'm realizing this design has a problem: if a contract wants to make use of this feature in an %on-push arm, they need to somehow keep a list of contract IDs which they know to correctly implement the fungible %push. In the current implementation, a malicious contract can *easily* produce an %on-push call, and the receiving contract has no way to verify if the tokens were actually given to them.

dr-frmr (changes requested):

lib/zig/contracts/lib/fungible.hoon
dr-frmr: The way to fix this is to actually make use of the %pull process inside of a %push. The only way I can think of for a receiving contract to reliably verify their payment is to have them %pull tokens as a continuation -- there's no straightforward way to prove that you've just added some number to their token balance. So, a %push should set an allowance, issue a continuation call, and expect the receiving contract to immediately %pull that amount.

lib/zig/contracts/lib/fungible.hoon
tadad: not the way I have it right now because you have to compare the current nonce `0` to the incoming nonce (has to also be `0`), and then increment the current nonce. You could do it the other way where you compare `+(current-nonce)` to `incoming-nonce`. Either way would work but my way indexes from zero rather than 1 which I like.

lib/zig/contracts/lib/fungible.hoon
tadad: right - this is also a problem with ERC677. In practice what you have is usually just a whitelist of tokens that you will accept, and that you know are not malicious. E.g. Chainlink says the `%on-push` has to come from the LINK token, and nothing else is a valid payment. In the case of something like uniswap where you want to use the `%on-push` interface for all token pools, abusing the `on-push` is just creating a scam token. However, this is still not ideal, and fee-on-transfer tokens would complicate this.

Another pattern we could use is the `approveAndCall` which is discussed in the 677 post [here](https://github.com/ethereum/EIPs/issues/677#issuecomment-317118992). This means that the receiver contract has to `%take` the tokens, which fixes the problem you brought up - the receiver contract can verify that it got the tokens. It is also more flexible in that the contract can calculate the exact amount of tokens it needs, and only `%take` some of them.

Personally I like the `approveAndCall` pattern a lot - seems like a safe and ideal way to call contracts. If you're on board I can rewrite the `%push`/`%on-push` to match the `approveAndCall` pattern

dr-frmr: Yep, approveAndCall looks exactly right. Funny to see them having basically the same conversation we are now, 5 years ago.

tadad: just changed `%push` to be an `approveAndCall` implementation instead of a `transferAndCall` implementation

dr-frmr:
Sadly just realized that this change to the wallet's `token-account` type will lead to a modified state type for `wallet.hoon`.

You know what that means... `%state-1`

dr-frmr:
> Sadly just realized that this change to the wallet's `token-account` type will lead to a modified state type for `wallet.hoon`.

As well as needing a `%state-1` to migrate wallets from current token account type, we'll also need to keep a version of the *old* fungible token type in order to support the currently-deployed testnet tokens :/

tadad:
> > Sadly just realized that this change to the wallet's `token-account` type will lead to a modified state type for `wallet.hoon`.
> 
> As well as needing a `%state-1` to migrate wallets from current token account type, we'll also need to keep a version of the _old_ fungible token type in order to support

God damn, I'm seething

dr-frmr:
I know, it sucks :(

When we're ready to exit testnet world we can shut down all our old stuff and set up a new distribution. At that point we can wash away all the backwards compatibility and state transitions like tears in rain.

dr-frmr (approved):

PR #190, opened 2022-10-05
wallet: sign as scry

hosted-fornet:
I think this is a way better interface for signing messages using keys stored in the wallet. I'm using this functionality in a subscription-gating project over [here](https://github.com/hosted-fornet/whitelist/pull/3). I like the `%sign-message` path best, but left the other in as well in case there is something I was missing about why typed messages are so special -- it seems to me that the serialization of the message should be up to the user, not the wallet to decide.

What do you think? @dr-frmr @tadad 

For reference, the previous way to do this is a poke [here](https://github.com/uqbar-dao/ziggurat/blob/cd771fb1b79acbccb9c4ae0a62daed8e83bc8088/app/wallet.hoon#L189-L202), followed by a scry to look at `signatures` stored in state. Unless there is some reason to keep the signatures around, this poke should be removed.

### DIFF
--- a/app/uqbar.hoon
+++ b/app/uqbar.hoon
@@ -71,20 +71,23 @@
     |=  =path
     |^  ^-  (quip card _this)
     ?>  =(src.bowl our.bowl)
-    :_  this
     ?+    -.path  !!
-        %track  ~
+        %track  `this
         %wallet
-      ::  must be of the form, e.g.,
-      ::   /wallet/[requesting-app-name]/[*-updates]
-      ?.  ?=([%wallet @ @ ~] path)  ~
+      ?.  ?=([%wallet @ @ ~] path)
+        ~|  "%uqbar: wallet subscriptions must be of form:"
+        ~|  " /wallet/[requesting-app-name]/[*-updates]"
+        !!
+      :_  this
       (watch-wallet /[i.path]/[i.t.path] t.t.path)
     ::
         %indexer
-      ::  must be of the form, e.g.,
-      ::   /indexer/[requesting-app-name]/grain/[town-id]/[grain-id]
-      ?.  ?=([%indexer @ @ ^] path)  ~
+      ?.  ?=([%indexer @ @ ^] path)
+        ~|  "%uqbar: indexer subscriptions must be of form:"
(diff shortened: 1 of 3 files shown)


dr-frmr:
I think this is a great improvement

hosted-fornet:
Added some bugfixes and cleanup.

dr-frmr (approved):

tadad:
yeah I think the main UX thing we want is the ETH [signTypedMessage](https://docs.metamask.io/guide/signing-data.html#a-brief-history) which displays exactly the information you are signing. Just signing an atom is really dangerous because you have no idea what is in the atom. Having it be like "you are signing a message that allows wallet A send B tokens to wallet C, which expires at timestamp D" is really helpful. So if we can build the frontend such that it is able to do that, then this change is welcome.

There's like a `+$typed-message` type I added a while ago, the goal of which was to mirror [EIP712](https://eips.ethereum.org/EIPS/eip-712)

hosted-fornet:
Thanks, this is good context. Given this information and a comment from dr-frmr on a dev post made today, I think we should revert this PR.

This PR makes it easier for apps to sign arbitrary messages, but ultimately that is bad for security. It seems we want the signing of a typed message to be very analogous to the signing of a transaction:
1. An app submits the signing request,
2. The wallet GUI prompts the user with information about the signing request and a yes/no button,
3. The app that requested the signing gets an update (via subscription?) with the signature or a `~` or some such.

This PR compresses that process into
1. An app submits signing request and receives signature back.

I think there is some work to do to make the `typed-message` real. The big benefit from having the message data be structured is that it can be displayed on the frontend. It looks like the way EIP712 does it is send the schema and the message together. The most obvious way to do this that comes to mind for me is to pass a `jold` along with the message. So maybe something like
```hoon
+$  typed-message  [domain=@ux =jold message=*]  ::  analogous to EIP-712
```

tadad:
yes I totally agree - right now the "signed message" doesn't really include the typehash. I thought about doing this somehow jamming the bunted hoon type, but just having it be json like EIP712 is good as well

hosted-fornet:
Yeah it would be nice if we could avoid using jolds and just use molds or something, but I suspect we will run into similar problems as led us to use jolds on-chain. Might be worth thinking about a bit more, though.

PR #194, opened 2022-10-11
Autogenerate grain ids

### LINKED ISSUE #193: Generate grain IDs automatically
When building a grain, we can actually generate the ID automatically based on the input values (lord/holder/town/salt). 

On the backend, adjust this behavior in the %add-to-state and %add-test-expectation pokes, such that you don't need to provide an ID field.

0x70b1a5:
#193


app/ziggurat.hoon
```diff
@@ -206,13 +206,14 @@
     ::
         %add-to-state
       ::  take text data input and ream to form data noun
-      =/  data-text  ;;(@t data.rice.act)
-      =.  data.rice.act
+      =/  rice  (fry-rice:smart lord.act holder.act town.act salt.act)
+      =/  data-text  ;;(@t data.rice)
```
dr-frmr, line 210: `fry-rice` produces a `@ux`, not a `rice`. So `data.rice` won't work here. You need to still bring in all the rice data via the poke, and produce the ID like you're doing.

app/ziggurat.hoon
```diff
@@ -248,7 +249,7 @@
       ::  add/replace expected rice output
       ?~  current=(~(get by tests.project) id.act)
         ~|("%ziggurat: test does not exist" !!)
-      =*  rice  expected.act
+      =*  rice  (fry-rice:smart lord.act holder.act town.act salt.act)
```
dr-frmr, line 252: Same issue here

sur/ziggurat.hoon
```diff
@@ -60,11 +60,11 @@
           [%save-file name=@t text=@t]  ::  generates new file or overwrites existing
           [%delete-file name=@t]
           ::
-          [%add-to-state =rice:smart]
+          [%add-to-state lord=id:smart holder=id:smart town=id:smart salt=@]
```
dr-frmr, line 63: The poke is missing some of the fields needed to create a `rice`. There are only the fields needed to generate the hash -- still need `label` and `data`

dr-frmr (changes requested):

app/ziggurat.hoon
```diff
@@ -205,17 +205,19 @@
     ::  local chain state management
     ::
         %add-to-state
+      =/  id  (fry-rice:smart lord.act holder.act town-id.act salt.act)
+      =/  rice  [salt=salt.act label=label.act data=data.act id=id lord=lord.act holder=holder.act town-id=town-id.act]
```
dr-frmr, line 209: Can either break this line into multiple lines (to stay below 80char line width) or preferably remove the assigned faces and instead impose them by doing `=/  =rice  [salt.act label.act data.act id lord.act holder.act town-id.act]`

app/ziggurat.hoon
```diff
@@ -205,17 +205,19 @@
...
       ::  take text data input and ream to form data noun
-      =/  data-text  ;;(@t data.rice.act)
-      =.  data.rice.act
+      =/  data-text  ;;(@t data.rice)
+      =.  data.rice
```
dr-frmr, line 212: To make this less repetitive, can move these lines above 209 and generate the proper data for the rice there. Can get rid of the `=.` if we just put the output of `text-to-zebra-noun` in the rice to begin with.

mar/ziggurat/contract-action.hoon
```diff
@@ -39,7 +39,6 @@
       :~  [%salt ni]
           [%label (se %tas)]
           [%data so]  ::  note: reaming to form data noun
-          [%id (se %ux)]
```
dr-frmr, line 42: I just realized that it's possible that this `parse-rice` arm might be used elsewhere to parse rice with IDs. It might make sense to create a variant that still does, while naming this version something like `parse-rice-without-id`

app/ziggurat.hoon
```diff
@@ -248,7 +250,8 @@
       ::  add/replace expected rice output
       ?~  current=(~(get by tests.project) id.act)
         ~|("%ziggurat: test does not exist" !!)
-      =*  rice  expected.act
+      =/  id  (fry-rice:smart lord.act holder.act town-id.act salt.act)
+      =/  rice  [salt=salt.act label=label.act data=data.act id=id lord=lord.act holder=holder.act town-id=town-id.act]
```
dr-frmr, line 254: Same as above re: line width and faces

app/ziggurat.hoon
0x70b1a5: Moving to something like this: 
```hoon
    %add-to-state
      =/  data-text  ;;(@t data.act)
      =/  id  (fry-rice:smart lord.act holder.act town-id.act salt.act)
      =/  rice  :*
          id 
          salt.act 
          lord.act 
          label.act 
          holder.act 
          town-id.act
          data=(text-to-zebra-noun data-text smart-lib-vase)
        ==
```
Do you know if assigning faces is necessary?

app/ziggurat.hoon
```diff
@@ -251,7 +257,15 @@
...
-      =/  rice  [salt=salt.act label=label.act data=data.act id=id lord=lord.act holder=holder.act town-id=town-id.act]
+      =/  rice  :*  
+            salt=salt.act
+            label=label.act
+            data=data.act
+            id=id
+            lord=lord.act
+            holder=holder.act
+            town-id=town-id.act
+          ==
```
dr-frmr, line 268: Formatting for this should align with standard backstep notation:
```
=/  rice
  :*  salt=salt.act
      label=label.act
      data=data.act
      id=id
      lord=lord.act
      holder=holder.act
      town-id=town-id.act
  ==
```

dr-frmr: Can keep the `:*` on the same line as `=/  rice` but then need to indent all the entries in the list to match

dr-frmr (approved):

dr-frmr:
Ok looks good!

PR #197, opened 2022-10-12
contracts: move to top-level

hosted-fornet:
Rearrange the repo to move contracts out of `lib/zig/` to top-level. Contracts are analogous to Gall apps, so should live at top-level. Also important because we want to merge `%ziggurat` Gall app/contract projects, and so contracts must be easy to find, rather than buried multiple levels deep in the directory tree.

FYI @tadad @willbach @0x70b1a5


tadad:
maybe `con/` instead of `contracts/` for the 3 letter aesthetic

dr-frmr (approved):
LGTM

PR #206, opened 2022-10-13
Add size limit for nouns generated by ziggurat's test/data compilation

dr-frmr:


dr-frmr:
@0x70b1a5 we should improve the UX for this by having some kind of error surfaced on-crash, LMK if that needs additional changes here. The issue was that storing a mold in the resulting noun made it too large to send over HTTP or even show it in dojo.

0x70b1a5 (approved):

0x70b1a5:
@dr-frmr looks good, i'll add the alert.

PR #209, opened 2022-10-16
mill: virtualize build process so as never to crash

dr-frmr:
Need to check if this is correct, but the idea here is to fix a bug where a malformed contract nock gets run in `mill` and crashes the process. We experienced this on the testnet recently, so it's important to get a fix out there soon. However, this does lower the performance of contract execution a bit -- should figure out if there's a way to reduce the load here, possibly by storing contracts as a single noun rather than a battery/payload pair.


dr-frmr:
Closing this since it will happen in `next/smart`

PR #212, opened 2022-10-17
Great Renaming

dr-frmr:
Going to be a lot of breaking changes coming out of this one

NOTES:
- also changed folder structure, grouped all sur and lib files in `zig` folders for convenience
- made a couple upgrades to the contract compilation process
- pokes/subs not exhaustively tested so possibility of broken ones
- some code might be ugly now (>80c lines), can upgrade style piecemeal in `next/smart`
- `mill.hoon` is now `engine.hoon` and is pending a large refactor, so is ugly atm.

Name mapping (main ones, some other small ones):
```
egg  ->  transaction
yolk  ->  calldata

grain  ->  item
rice  ->  data
wheat  ->  pact

mill  ->  engine
mill-all  ->  run

granary  ->  state
populace  ->  nonces
land  ->  chain
town  ->  shard

basket  ->  mempool
carton  ->  memlist 

(some conceptual changes here as well)
rooster  ->  diff
hen  ->  call
chick  ->  (quip call diff)
```


dr-frmr:
Renaming is complete.

Not sure how we should approach review on this, but I'm certain of two things:
- there are lots of references to old names I missed that aren't causing compilation issues
- the frontend subs/pokes/scries are probably mostly broken

Don't think we have to fix either of those things in this PR, but rather piecemeal as we build on `next/smart`.

I haven't tested the frontend at all in conjunction with this, but I've ran the fakechain startup commands and run a bunch of transactions. The only contract I've fixed up / recompiled is zigs.hoon. I'll do the other ones after we get this merged so other people can start building on the new names right away.

If there are any naming issues please LMK. The one I like least is probably `data`, so if anyone has a stroke of genius there that'd be great. Don't want to have to go through this again so let's get the names locked in here.

dr-frmr:
@hosted-fornet can you double-check my merging in of your state-transition-removal work from `next/smart` to here? I think it's right but you'd know best

hosted-fornet:
For consistency I'd recommend combining `lib/wallet/util.hoon` and `lib/wallet/parsing.hoon` into a single file and moving into `lib/zig` with other libraries: `lib/zig/wallet.hoon`.

hosted-fornet:
Tests: update this PR or another PR?

hosted-fornet:
Basic functionality works. Still work to be done to get, e.g., tests up to speed. If we want to do that in a separate PR, then this PR is good to go. As @dr-frmr says above, I expect there will be lots of little things that need fixing moving forward, but sends for zigs and NFTs work, indexer works, wallet works :+1:

dr-frmr:
Alright let's do it. Tests and more to come next week.

hosted-fornet:
Nice work on this

PR #214, opened 2022-10-18
indexer: rip out subscription logic

hosted-fornet:
Rip out subcription logic from %indexer. Update %wallet to avoid using `/id` subscription path -- which no longer exists.

`/batch-order/[town-id]` (or `/json/batch-order/[town-id]`) is now main subscription path. `/batch-order` DOES reply on watch with entire `batch-order` history. New model is to subscribe to `/batch-order`, and when receive a %fact, scry for the data you need in the new batch.

Please take a look in particular at %wallet changes: I am less familiar with this code. I tried to keep functionality the same while moving to the new model.

FYI @willbach @0x70b1a5 : you will have to update the frontend if you were relying on subscriptions other than `/batch-order` / `/json/batch-order`. Happy to help here.

### DIFF
--- a/app/indexer.hoon
+++ b/app/indexer.hoon
@@ -81,63 +81,16 @@
 ::
 ::    ## Subscription paths
 ::
-::    Subscriptions paths must be appended with a unique
-::    subscription identifier.
-::    The %uqbar app does this automatically.
-::    One recommended way to do this is to append
-::    ```
-::    (scot %ux (cut 5 [0 1] eny.bowl))
-::    ```
-::    to the path.`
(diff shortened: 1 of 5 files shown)


hosted-fornet:
It can wait till after.

dr-frmr:
Renaming PR is set so you could retarget this to there if you want to start building.

hosted-fornet:
FYI @dr-frmr : merged and changed base to Great Renaming branch.

dr-frmr (approved):
Is it possible to simplify indexer types as well? Perhaps simplify the update types now that we're only scrying?

PR #217, opened 2022-10-20
new poke for approving cors domains

0x70b1a5:


0x70b1a5:
Hang on, this doesn't solve our problem. 

When WE approve CORS domains, that lets _incoming_ requests access _our_ site (in this case the page served by the user's ship), but we have no ability to make _Github_ approve our site's _outgoing_ requests. Man, this is confusing. 

Might still be useful to have this poke in there, but we will need something else to fix the zip download.

app/ziggurat.hoon
```diff
@@ -499,6 +499,19 @@
         /(scot %p our.bowl)/(scot %tas project.act)/(scot %da now.bowl)
       :-  (make-app-update project.act app-project)^~
       state(projects (~(put by projects) project.act %|^app-project))
+    ::
+        %approve-cors-domain
+      ::  for internal use -- required to download github repo zips on frontend
+      :_  state
+      :~  :*  %pass
```
hosted-fornet, line 506: Style nit: I would recommend substituting the wrapping `:~  ...  ==` with a `:_  ~`: it unindents your `:*` and, more importantly, it reduces your use of arbitrary-length runes. A general rule: if you can use a fixed-length rune, like `:+` or, in this situation, `:_`, use that over an arbitrary-length rune like `:*`. In doing so, you are making more explicit constraints on your code, similar to declaring types. Get the compiler to work for you.

hosted-fornet (approved):
LGTM :+1:

PR #222, opened 2022-10-24
fix create-new-project mar

0x70b1a5:
This contains only the minimum necessary to create a new project from a github repository. Files without known marks are not handled correctly, but I think it is expedient to get these changes to ziggurat marks/parsers/etc into `next/smart` before anybody else does work.


dr-frmr (approved):
excellent, ty

PR #223, opened 2022-10-24
Some mar file updates for frontend

dr-frmr:


lib/zig/indexer.hoon
```diff
@@ -279,7 +279,7 @@
         ==
       ::  wheat
       :~  [%is-data %b %|]
-          [%cont (numb (jam code.p.item))]
+          [%cont (numb 0)]
```
dr-frmr, line 282: Note @hosted-fornet this process was causing serious delay when looking at a contract on frontend. We should come up with an alternative mechanism to display the contract's nock code, preferably pretty-printed, preferably once (given that deployed contracts will likely be immutable)

hosted-fornet: Did you confirm this speeds up the page spinner? I had tracked it down to a different source, namely a `/hash` query for the zigs contract, which gave hundreds of batches worth of history of all `/source`d zigs accounts. That said, I agree with this change.

dr-frmr: Yeah, it worked for me

hosted-fornet: Cool :+1:

hosted-fornet: Tested on a fakeship testnet? The thing I was seeing will only be slow with many batches and many users.

dr-frmr: Yes, this just solves slowness of one kind. We still need a slowness solution for the thing you're talking about.

PR #225, opened 2022-10-25
engine.hoon refactor

dr-frmr:
* refactored `engine` for readability, flow, correctness
* wrote a new test suite with the goal of proving correctness
* broke out engine flags into two, `sigs-on` and `hints-on`, for testing and PoA usage
* for now, removed all passing of zink hints in engine -- can easily re-introduce when they are ready

TODO:
* [x] add processed transaction list to output type
* [x] write more tests, cover reads, continuations, and zigs payments more thoroughly
* [x] write multi-transaction tests
* [x] document with comments a bit better


dr-frmr:
Still more aspects of the engine that deserve testing but at this point I'm confident that the refactor has fewer bugs than the original. Feeling very good about this, will continue to write tests (and documentation...) but merging now

PR #238, opened 2022-10-28
Make thread to nuke all apps on desk

dr-frmr:


ted/upgrade.hoon
```diff
@@ -0,0 +1,28 @@
+/-  spider
+/+  strandio
+=,  strand=strand:spider
+::
+^-  thread:spider
+|=  arg=vase
+=/  m  (strand ,vase)
+^-  form:m
+=/  our=ship  (need !<((unit ship) arg))
```
hosted-fornet, line 9: Assuming we only ever want this to be `our.bowl`, an alternative way to do this would be to

```hoon
;<  our=ship  bind:m  get-our:strandio
```

which means no args to the thread are needed and it would just be run from Dojo as

```
-zig!upgrade
```

hosted-fornet (approved):
Added a nit. Otherwise LGTM and code runs as expected :+1:

ted/upgrade.hoon
dr-frmr: Ty, I am not familiar with the strandio library. I will do this

dr-frmr:
Ok, in order to upgrade from `stable-release-0.0.4` to what's being built in `next/smart`, we'll need all our users to use this thread. We'll need to spin up a private internal testnet where we can test this migration. I think the flow can work like this:

(sidenote: let's start just calling it `release` since they're not that stable yet, lol)

- users are watching ~dister ship that's on `release`
- ~dister ship upgrades to `next/smart`, sending out an incompatible update to users
- at any point after this, users instructed to run `-zig!upgrade` will nuke their apps in %zig desk
- this will either automatically trigger the queued update, or they'll have to run `|install` manually
- after which, they'll be in the new system with no issue

Simultaneously, we'll have to nuke the testnet state on ~bacdun. This will be more fraught, as we don't want upgrade incompatibilities between the testnet and ships watching it -- this is part of what caused our last breaking bug. I think a good solution might be to first put out a software update that removes the "resub on kick" behavior from key apps, then kick all subscribers from update paths on ~bacdun. We can reintroduce the "resub on kick" in `next/smart`, but until users get that update they'll be passively disconnected from the testnet. After ~bacdun is isolated, we can separately nuke all it's apps and install the upgrade, then start the new testnet so it's ready for upgraded users to subscribe to.

ted/upgrade.hoon
hosted-fornet: strandio is great! Threads are fun

dr-frmr:
Going to merge the file built here into both `master` and `next/smart` from #249

PR #247, opened 2022-11-03
assume new files have their own mars

0x70b1a5:
Here is a summary of the issue this commit attempts to address.

- Users may create files of any type.
- The parent commit assumes that unmarked files are %hoon.
- As a result, parent commit crashes when new files are invalid hoon, e.g., `''`.

On the frontend side of things, when a user tries to create a file with an unknown extension, we currently create a mar for that extension based off of `/mar/txt/hoon`. Then, we create the file using the mar we just created. That's what this commit is doing.

I invite feedback, as I doubt this is the best way to do this.


lib/zig/ziggurat.hoon
```diff
@@ -99,8 +99,10 @@
   :_  ~  :+  file  %ins
   ?:  =(%hoon file-type)
     [%hoon !>(`@t`text)]
+  ?.  (~(has in (silt `(list @tas)`~[%ship %bill %kelvin %docket-0])) file-type)
```
0x70b1a5, line 102: i'm aware this isn't very 'hooney' given how heavy the line is. welcoming suggestions

dr-frmr: I don't think this line is necessary at all given that we go into a switch statement immediately afterwards. Why don't we put ```[`@tas`file-type !>(`@t`text)]``` as the default case in the `?+`? Then we just cover `[%ship %bill %kelvin %docket-0 %hoon]` as our switch cases.

dr-frmr: Also there's nothing wrong with the line itself, stylistically -- it's under 80 characters

0x70b1a5: oh, i guess you can just do that, huh. man!

0x70b1a5: @dr-frmr the issue is that if `text` is empty, we can't `ream` it

dr-frmr: Well, that's okay, because if someone is trying to save an empty .hoon file we shouldn't allow that. Can either just crash during this poke, or return some other effect to FE

0x70b1a5:
Sorry, I am not speaking clearly. What I mean is, txt has a broader
validity domain than Hoon. `''` is not valid Hoon, but it is valid txt.
Therefore if we ream any possible value of the type `hoon | txt` before
entering the switch, the code will crash even for some valid values of txt.
Maybe we could only ream known hoon types inside the switch?

lib/zig/ziggurat.hoon
hosted-fornet: > Sorry, I am not speaking clearly. What I mean is, txt has a broader validity domain than Hoon. `''` is not valid Hoon, but it is valid txt. Therefore if we ream any possible value of the type `hoon | txt` before entering the switch, the code will crash even for some valid values of txt. Maybe we could only ream known hoon types inside the switch?

This is a good point, but I don't think the code currently makes it clear that that is the case we are concerned about. One possible solution would be a comment. Another, that might be a little subtle, would be to do the change @dr-frmr recommends and turn [line 104](https://github.com/uqbar-dao/uqbar-core/blob/188edaab00dcd5edf5621675b383bc27163e033c/lib/zig/ziggurat.hoon#L104) into a `=*`, which will delay computing the `+ream` until it is needed -- and if it is not needed, as in the default `?+` case, then it will not be computed.

hosted-fornet: I don't know that the `=*` makes it particularly clear what is going on either -- but I think it would make the switch statement work as expected.

hosted-fornet: So perhaps the best solution is a comment + current code or comment + `=*` and `?+` default -> ```[`@tas`file-type ...]```

hosted-fornet (approved):
Code comment (or minor change) requested -- otherwise LGTM. (I did not test).

lib/zig/ziggurat.hoon
dr-frmr: I guess I’m confused why we can’t move the ream into the switch statement where it won’t be executed on non-hoon files, but I get the gist of this. Lgtm

hosted-fornet: > I guess I’m confused why we can’t move the ream into the switch statement where it won’t be executed on non-hoon files, but I get the gist of this. Lgtm

Could do that as well. Can that be done without writing the `+ream` one time for each branch, though?

app/ziggurat.hoon
```diff
@@ -142,7 +142,7 @@
       =.  user-files.project
         (~(put in user-files.project) file.act)
       :_  state(projects (~(put by projects) project.act project))
-      :+  (make-save-hoon [project file text]:act)
+      :+  (make-save-file [project file text]:act)
```
hosted-fornet, line 145: Good change

app/ziggurat.hoon
```diff
@@ -535,23 +535,29 @@
     =/  des=@ta    i.t.t.path
     =/  pat=^path  `^path`t.t.t.path
     =/  pre=^path  /(scot %p our.bowl)/(scot %tas des)/(scot %da now.bowl)
-    =/  res        .^(? %cu (weld pre pat))
-    ``json+!>(`json`[%b res])
+    =-  ``json+!>(`json`[%b -])
+    .^(? %cu (weld pre pat))
```
hosted-fornet, line 539: nit: would prefer on one line like
```hoon
``json+!>(`json`[%b .^(? %cu (weld pre pat))])
```

downside of doing on one line is you get lisp parens/brackets.

hosted-fornet:
Tested and `%read-file` worked as expected on `.bill`, `.hoon`, and `.jam` (once I added symlinks to `sur/dice.hoon` and `lib/tiny.hoon`). Didn't test `+make-save-file`. I assume this is working with the frontend?

Other than some comments above (nits optional, others preferred), LGTM. Nice work! :+1:

0x70b1a5:
Yeah, +make-save-file is working at least for making non-hoon files with custom marks.

hosted-fornet:
FYI @0x70b1a5 not sure if you intended to push more commits here since I only see one numbered commit here labeled `nit #1`

PR #248, opened 2022-11-04
Update item in fake chain state

willbach:
Should I pull out most of the logic in this poke since it's largely duplicative of `%add-to-state`?


dr-frmr:
@willbach, I'd say yes, pulling out the logic and putting it somewhere once would be good here. Also, if we've got `%add-to-state` and `%delete-from-state`, I feel like we should rename them to match this naming convention. `%add-item`, `%edit-item`, `%delete-item`. Remember that new or changed pokes also need to be modified in the associated `mar` file

willbach:
Confirmed that create, edit, and delete pokes are working end-to-end.

PR #255, opened 2022-11-09
engine: fix validation check for issued pacts

dr-frmr:
The line here was causing immutable contracts deployed through publish.hoon to fail validation. This was because we were checking the pact ID against the source, which was the ID of publish.hoon, but should be `0x0` in immutable contracts! Easy fix.


hosted-fornet (approved):
LGTM

PR #256, opened 2022-11-10
pyre,ziggurat: make tweaks to get working fast

hosted-fornet:
Some fixes to %pyre and tweaks to %ziggurat to get things running fast. Removing the vane-threads seems to have sped things up significantly.

Got rid of ~zod and ~dun pinging by changing desk.ship and indexer `+on-init`, respectively: this saves a lot of time. However, we should figure out how to better deal with incorrectly configured desk.ship files so that our users don't have a huge slowdown if they copy in one that doesn't point to ~nec. A sideeffect of pointing at ~nec is that installing from pill on ~bud causes a remote install from ~nec. Here the ames packet size change Philip recommended does serious work: ~11s -> ~3s run time. Ideally this should not happen: I think we should not distribute a pill like I am doing with %base + %zig, but instead just %base and copy files in. Once that is working, the ~nec remote install and desk.ship problem should go away.

One thing I looked at but didn't seem to make any difference was bouncing the %pyre `+on-agent` calls through a self-poke to do something like [referenced here in aqua-vane-thread](https://github.com/urbit/urbit/blob/master/pkg/arvo/lib/aqua-vane-thread.hoon#L49-L53). This comment may just be out of date: I didn't see any difference in behavior when doing vs not doing this. However, if we see out of order subscriptions, make a mental note to try this as a fix.


hosted-fornet:
@tadad if you find this PR good, and you don't mind merging it into your branch, let me know. I think we should start merging our work, so upstream branches don't have to merge downstream ones all the time. Hopefully a release is cut soon so we can then just merge into next/smart (or possibly we should merge into an entirely separate branch? @dr-frmr )

hosted-fornet:
@tadad : PS: feel free to merge this yourself if you approve.

dr-frmr:
Do you guys want a `next/pyro` to target? It should definitely be stacked on top of `next/smart` but that should get merged into master by the end of this week.

tadad:
yeah these branches are a little disorganized so let's start merging

PR #257, opened 2022-11-10
Add contract field to wallet's `asset-metadata`

dr-frmr:
While building pokur, realized that wallet needs to provide this information to other apps somehow. I also added a scry path where apps can get token metadata so they can use it to create transactions.

### DIFF
--- a/app/wallet.hoon
+++ b/app/wallet.hoon
@@ -562,8 +562,12 @@
     =/  pub  (slav %ux i.t.t.path)
     =/  our=(map @ux [transaction:smart supported-actions])
       (~(gut by pending-store) pub ~)
-    ::
     ``noun+!>(`(map @ux [transaction:smart supported-actions])`our)
+  ::
+      [%metadata @ ~]
+    ::  return (unit) specific metadata from our store
+    =/  found  (~(get by metadata-store) (slav %ux i.t.t.path))
+    ``noun+!>(`(unit asset-metadata)`found)
   ==
 ::
 ++  on-leave  on-leave:def
--- a/sur/zig/wallet.hoon
+++ b/sur/zig/wallet.hoon
@@ -15,8 +15,8 @@
 ::
 +$  metadata-store  (map id:smart asset-metadata)
 +$  asset-metadata
-  $%  [%token town=@ux token-metadata]
-      [%nft town=@ux nft-metadata]
+  $%  [%token town=@ux contract=id:smart token-metadata]
+      [%nft town=@ux contract=id:smart nft-metadata]
   ==
 ::
 +$  unfinished-transaction-store
--- a/lib/zig/wallet.hoon
+++ b/lib/zig/wallet.hoon
@@ -149,8 +149,8 @@
     ~
   ?.  ?=(%& -.u.g)  ~
   ?+  token-type  ~
-    %token  `[%token town ;;(token-metadata noun.p.u.g)]
-    %nft    `[%nft town ;;(nft-metadata noun.p.u.g)]
+    %token  `[%token town source.p.u.g ;;(token-metadata noun.p.u.g)]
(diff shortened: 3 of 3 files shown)


app/wallet.hoon
```diff
@@ -469,6 +469,62 @@
   ?.  =(%x -.path)  ~
   =,  format
   ?+    +.path  (on-peek:def path)
+  ::
+  ::  noun scries, for other apps
+  ::
+      [%keys ~]
+    ``wallet-scry+!>(`wallet-scry`[%addresses ~(key by keys.state)])
```
hosted-fornet, line 476: The difference between endpoint and output label is weird here since they are the same elsewhere (except for `%transaction`). Recommend either changing endpoint to `%addresses` or label to `%keys` so they match.

app/wallet.hoon
```diff
@@ -469,6 +469,62 @@
...
+  ::
+      [%signatures ~]
+    ``wallet-scry+!>(`wallet-scry`[%signatures signatures.state])
+  ::
+      [%metadata @ ~]
+    ::  return specific metadata from our store
+    :^  ~  ~  %wallet-scry
+    !>  ^-  wallet-scry
+    ?~  found=(~(get by metadata-store) (slav %ux i.t.t.path))
+      ~
```
hosted-fornet, line 495: Here and other null cases below: should be `[~ ~]`. Semantics of `~` is "scry path couldn't be resolved" and `[~ ~]` is "no data", see https://developers.urbit.org/guides/core/app-school/10-scry#handling-scries

app/wallet.hoon
```diff
@@ -562,7 +603,6 @@
     =/  pub  (slav %ux i.t.t.path)
```
hosted-fornet, line 603: Move `%pending-noun` up to noun section.

sur/zig/wallet.hoon
```diff
@@ -50,13 +50,38 @@
...
   $%  [%new-book tokens=(map pub=id:smart =book)]
       [%new-metadata metadata=metadata-store]
-      [%finished-tx hash=@ux =transaction:smart action=supported-actions =output:eng]
       [%tx-status hash=@ux =transaction:smart action=supported-actions]
+      $:  %finished-tx
+          hash=@ux
+          =transaction:smart
+          action=supported-actions
+          =output:eng
+      ==
```
hosted-fornet, line 84: Is there a reason `wallet-scry` and `wallet-update` cannot be combined? If not, they should be combined.

hosted-fornet: Could then combine `%finished-transaction` with `%finished-tx`

hosted-fornet (approved):
Looks mostly good, left some comments. I'll make an issue for non-immediate stuff that would be nice to have. I did not test.

app/wallet.hoon
dr-frmr: true, i will change scry endpoint to addresses since that's more accurate

app/wallet.hoon
dr-frmr: I'm wrapping the `~` inside the wallet scry though, as in, returning an empty result

dr-frmr: Which I'm doing because I don't want the scry to crash if the thing isn't found

sur/zig/wallet.hoon
dr-frmr: I really like having the types that are purely for frontend use separate from those which are used in other apps, but I think I should rename `wallet-scry` to `wallet-update` in order to match indexer, and then I'll rename the frontend update type accordingly

app/wallet.hoon
hosted-fornet: Oh right, my mistake.

dr-frmr:
Fixed issues above, will merge and deal with issue #264 in future

PR #258, opened 2022-11-10
pyro: snapshots

hosted-fornet:
WIP work on %pyro snapshots

Plan is for %ziggurat to store snapshots with path
```hoon
/[project-name]/[test-id]/[step]
```

If I can get test steps discretized here I'll add snapshotting per test-step here as well.

Also want to explore the possibility of `+jam`ing starting state for ships and `+cue`ing it into `+on-init` state of %pyro to avoid having to do the long startup for each ship.


hosted-fornet:
I played around a bit with exporting state like `fresh-piers` and `snap`s in order to speed up loading. Ran into issues with jet mismatches. If we can figure out jet mismatching, this can be a good way to speed up loads, but am going to set aside for now.

PR #259, opened 2022-11-10
next/pyro

hosted-fornet:
WIP


hosted-fornet:
Despite dev-suite being spun out, this PR should still be merged, as it has some changes to indexer, wallet, etc. It also removes a bunch of stuff since dev-suite is now independent. Please take a look and merge when you have time @dr-frmr

hosted-fornet:
https://github.com/uqbar-dao/uqbar-core/blob/d6ca4143ddc5a7b8695314469840d4767d76ea0b/app/indexer.hoon#L158-L159

These lines of %indexer need to be changed to ~nec and ~nec for use with dev-suite. Because we install this %zig desk by default on virtualships, we need to figure out a better way to assign these values before we release the dev-suite. It will not work with these values assigned to bacdun and whatever, and the livenet will not work with them assigned to ~nec and ~nec. For now, we will need to replace these lines while developing dev-suite @dr-frmr  @tadad @0x70b1a5

sur/zig/subscriber.hoon
```diff
@@ -0,0 +1,9 @@
+|%
++$  subscriber-action
+  $%  [%sub =ship app=@tas =path]
+      [%unsub =ship app=@tas =path]
+      [%clear =ship app=@tas =path]
+  ==
++$  signs  (map [=ship app=@tas =path] (set sign:agent:gall))
++$  facts  (map [=ship app=@tas =path] (set @t))
+--
```
dr-frmr, line 9: Is this file (`subscriber.hoon`) relevant to `uqbar-core`? It looks like something for pyro/ziggurat, not sure.

zig
```diff
@@ -0,0 +1 @@
+../dev-suite/zig
```
dr-frmr, line 1: Is this symlink necessary? If so, what's it used for? I kinda don't like that one might need to have `dev-suite` and `uqbar-core` both cloned in order to work on core

dr-frmr:
> These lines of %indexer need to be changed to ~nec and ~nec for use with dev-suite. Because we install this %zig desk by default on virtualships, we need to figure out a better way to assign these values before we release the dev-suite. It will not work with these values assigned to bacdun and whatever, and the livenet will not work with them assigned to ~nec and ~nec. For now, we will need to replace these lines while developing dev-suite @dr-frmr @tadad @0x70b1a5

Yeah, this is fine. Let's change the default to `~nec` and just swap that value out when we do deployments to `~bacdun`, this is already what we need to do in other places (sequencer whitelist, pokur host, others..)

dr-frmr:
@hosted-fornet I just caught `next/smart` up to `master` which introduced some small conflicts -- nothing bad.

sur/zig/subscriber.hoon
dr-frmr: Is the `subscriber` app supposed to be in this repo?

tadad: Yeah it does. It's kind of a dumb app, and I like the idea of getting rid of it

tadad:
> https://github.com/uqbar-dao/uqbar-core/blob/d6ca4143ddc5a7b8695314469840d4767d76ea0b/app/indexer.hoon#L158-L159
> 
> These lines of %indexer need to be changed to ~nec and ~nec for use with dev-suite. Because we install this %zig desk by default on virtualships, we need to figure out a better way to assign these values before we release the dev-suite. It will not work with these values assigned to bacdun and whatever, and the livenet will not work with them assigned to ~nec and ~nec. For now, we will need to replace these lines while developing dev-suite @dr-frmr @tadad @0x70b1a5

Has anyone discovered a better pattern for this? Like adding a poke to set the ships instead of hardcoding? This is a common thing that makes a lot of %pyro stuff break because it tries to contact a ship that isn't working

con/escrow.hoon
```diff
@@ -0,0 +1,194 @@
+::
+::  token escrow contract for managing pokur games
+::  (but general enough to be used for anything?)
+::
+/+  *zig-sys-smart
```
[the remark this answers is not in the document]
hosted-fornet: Hmm, I think I have a solution; I'll remove them here.

sur/zig/subscriber.hoon
tadad: deleted

hosted-fornet: %subscriber needs to be run on virtualships by default -- we assume it is running in %ziggurat so we can, e.g., extract +dbug and format it nicely. This may change with typed scries. So the answer is: for now yes; once typed scries are integrated with %ziggurat, probably no.

hosted-fornet: For more context, virtualships have uqbar-core installed on them by default, which is why it needs to live here, for now.

zig
hosted-fornet: If we want core to be importable to %ziggurat, we need a `zig/` directory with some of these files in it. We could maintain a copy of these files in both repos, as an alternative to symlinking.

hosted-fornet:
> > https://github.com/uqbar-dao/uqbar-core/blob/d6ca4143ddc5a7b8695314469840d4767d76ea0b/app/indexer.hoon#L158-L159
> > 
> > These lines of %indexer need to be changed to ~nec and ~nec for use with dev-suite. Because we install this %zig desk by default on virtualships, we need to figure out a better way to assign these values before we release the dev-suite. It will not work with these values assigned to bacdun and whatever, and the livenet will not work with them assigned to ~nec and ~nec. For now, we will need to replace these lines while developing dev-suite @dr-frmr @tadad @0x70b1a5
> 
> Has anyone discovered a better pattern for this? Like adding a poke to set the ships instead of hardcoding? This is a common thing that makes a lot of %pyro stuff break because it tries to contact a ship that isn't working

Having stuff hardcoded in on-init is a hack for sure, mostly done to make setup more automagic for users. The better solution is to have a FE setup box, so that when the user first opens up the core suite, they are prompted to select a %rollup and %sequencer host (perhaps with defaults of bacdun and dister... filled in the textboxes by default).

sur/zig/subscriber.hoon
hosted-fornet: Just so dev on FE can proceed without hiccups. We can keep this PR open till %subscriber can be removed so it isnt checked in to master. I'll add it back in in a moment.

dr-frmr:
It's fine to just keep all hardcodings to fakeship galaxies and manually swap out the on-init ships when we build deployments imo

dr-frmr:
Can the symlinks to `dev-suite` be removed now that the necessary ziggurat files are included? I'm fine keeping those in here, but not the pokur stuff (escrow contract)

hosted-fornet:
%indexer hardcodes to ~nec as of 7d213c6. d418833 removes dev-suite symlinks, in favor of bringing copies of those files in.

dr-frmr:
TY. approved

dr-frmr (approved):

hosted-fornet:
Whoops, forgot to update gen/sequencer/init/hoon now that escrow is gone.

hosted-fornet:
%pokur import working with https://github.com/dr-frmr/pokur/pull/29

PR #271, opened 2022-11-14
%share-address: map your @p to uqbar address(es)

dr-frmr:
name WIP

### DIFF
--- a/app/share-address.hoon
+++ b/app/share-address.hoon
@@ -1,0 +1,134 @@
+::  share-address [UQ| DAO]:
+::
+::  A simple agent for ships to broadcast the address they're using on Uqbar.
+::
+/-  *zig-share-address, w=zig-wallet
+/+  agentio,
+    dbug,
+    default-agent,
+    verb,
+    smart=zig-sys-smart
+::
+|%
++$  card  card:agent:gall
+--
+::
+=|  state-0
+=*  state  -
+::
+%-  agent:dbug
+%+  verb  |
+^-  agent:gall
+|_  =bowl:gall
++*  this  .
+    def   ~(. (default-agent this %|) bowl)
+    io    ~(. agentio bowl)
+::
+++  on-init  `this(mapping ~, known ~)
+++  on-save  !>(state)
+++  on-load
+  |=  old=vase
+  `this(state !<(state-0 old))
+::
+++  on-poke
+  |=  [=mark =vase]
+  ^-  (quip card _this)
+  ?.  ?=(%share-address-action mark)  (on-poke:def mark vase)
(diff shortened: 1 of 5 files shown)


dr-frmr:
Should this be its own repo/desk?

dr-frmr:
Perhaps this functionality is so integral to the wallet that it should simply be merged into the wallet agent itself?

dr-frmr:
I'm going to merge just the %wallet changes cause those are small, necessary, and break state, so I want to get them into `next/smart`. The rest can wait

hosted-fornet:
> Perhaps this functionality is so integral to the wallet that it should simply be merged into the wallet agent itself?

I think this could make sense. The main pro is %wallet becoming a one-stop-shop for a lot of Gall agent needs -- which is what we want. The main con is scope creep. Given recent discussions about Tlon learnings (such as advantages of monolithic agents), I think it probably makes sense to have this in %wallet.

Glancing at the code, I am wondering: does it make sense that a @p is one-to-one with an address? My intuition from ETH is that I might want many Uqbar addresses, and may only want some of them tied publicly to my @p.

hosted-fornet:
Also I like the use of agentio :+1:

hosted-fornet:
Individually specifying `@p`s that can view a given address seems clumsy, but would be the ultimate level of control. E.g. I might want my friends to know I have a given address, but no one else, or have a family-only address, or whatever. The problem becomes:

1. How to expose that interface to users in a non-clumsy way.
2. How to make it hard for users to accidentally expose "private" addresses unintentionally.

Although it starts to lose some of the niceness of this one-stop-shop, I imagine that some use cases --- like a "private" address -- might be better served by a protocol where the user passes in the signatures/pair of sigs to the app providing whatever service.

Does it make sense to have multiple protocols that are specialized for different things, like public vs private personas? Or should we try to serve all cases with one protocol? If multiple protocols, we will need tools that make it easy for app devs to implement the multiple protocols -- or the subset of protocols they desire.

Looking at the `known` `map` and `got-address` you have here, it makes me think the "private" protocol is actually a pretty trivial add on here: just have the user `poke` the `%wallet` on the provider's ship to populate with the user's "private" persona before doing whatever next part of the protocol.

dr-frmr:
I think what you're talking about here has potential to be a pretty rich social app, with obvious integrations with things like %pals, %contacts, and many apps and games. For that reason it might make sense to put it in its own desk with its own frontend. We might also separately want to have a 1:1 mapping stored in a contract *on-chain* as discussed in meeting today, which could fulfill the role this little agent has.

dr-frmr:
Closing this because solution will be a protocol on top of %social-graph

PR #274, opened 2022-11-15
indexer: make subscriptions resub on kick again, upgrade: make thread inert

dr-frmr:
For `release-0.0.6`: %indexer should return to resubbing on kick, and upgrade thread is no longer needed. Don't want people accidentally nuking all their app states, so we will make the thread inert until a future time when it may be needed again.


hosted-fornet (approved):
LGTM :+1:

PR #275, opened 2022-11-15
add docket file and jpg mar for globs

0x70b1a5:


hosted-fornet (approved):
Let's Get That Money :+1:

PR #286, opened 2022-11-18
zigs: remove `%give` and `%take` `to-account`

hosted-fornet:
Just putting some code up so we have something more concrete to discuss. If we decide we want to go forward with this I will make similar changes to fungible. This may also be better directed to next/smart?

```
>   "gas cost: 20.417"
```


hosted-fornet:
Please see PR #287 which I did not mess up by changing upstream branch.

PR #287, opened 2022-11-18
zigs: remove `%give` and `%take` `to-account`

hosted-fornet:
Just putting some code up so we have something more concrete to discuss. If we decide we want to go forward with this I will make similar changes to fungible. This may also be better directed to next/smart?

```
>   "gas cost: 20.417"
```


dr-frmr:
Been looking at this change and I think it's worth it. Also makes other contracts able to compose with tokens more easily: right now, contracts that want to execute %gives and %takes need to take in token account IDs and use them. This change simplifies not only the token standard, but every other contract that interacts with it.

dr-frmr:
@hosted-fornet are you cool if I start committing to this branch with the various changes needed to %wallet and other contracts?

hosted-fornet:
@dr-frmr Go for it: consider the PR yours.

dr-frmr:
This is ready for review @hosted-fornet @tadad -- surprisingly few breakages/dependencies were disrupted, thought there'd be more stuff. The breaking changes affect the AMM contract and pokur escrow contract, which I will change

tadad (approved):
looks good - code is a lot cleaner now

con/lib/fungible.hoon
```diff
@@ -126,53 +114,47 @@
     ^-  (quip call diff)
     =+  (need (scry-state from-account.act))
```
hosted-fornet, line 115: comparing to zigs.hoon, there is a line missing (zigs.hoon reads):

```hoon
    ?>  !=(to.act id.caller.context)
    =+  (need (scry-state from-account.act))
```

Should we be doing that assert here as well?

con/lib/fungible.hoon
```diff
@@ -126,53 +114,47 @@
...
     =/  giver  (husk account:sur - `this.context `id.caller.context)
```
hosted-fornet, line 116: nit: for consistency with zigs.hoon, would recommend changing this line to

```hoon
    =+  giver=(husk account:sur - `this.context `id.caller.context)
```

con/lib/fungible.hoon
```diff
@@ -229,21 +210,19 @@
     ?>  .=(nonce.act -)
```
hosted-fornet, line 210: nit: use form

```hoon
?> .=(nonce.act -)
```

hosted-fornet (commented):
Mostly nits, otherwise LGTM. Didn't test.

con/lib/fungible.hoon
dr-frmr: Ah, yes.

dr-frmr:
Merging. This PR also provides a bug fix to these token contracts that allowed people to send tokens to themselves and double their balances. Important to fix!

PR #291, opened 2022-11-21
readme: update a dojo command for building indexer molds

bacwyls:
looks like this file moved, but the path hasn't been update in the readme yet


hosted-fornet (approved):
Good catch, thanks for the fix!

PR #295, opened 2022-11-21
indexer: sequencer/rollup `+on-watch` -> ``this`

hosted-fornet:


app/sequencer.hoon
```diff
@@ -58,17 +58,14 @@
 ++  on-watch
   |=  =path
   ^-  (quip card _this)
-  ?.  =(%available status.state)
-    ~|("%sequencer: error: got watch while not active" !!)
   ::  open: anyone can watch
+  :: ?.  =(%available status.state)
+  ::   ~|("%sequencer: error: got watch while not active" !!)
```
hosted-fornet, line 63: Please note this check is now commented to avoid breaking fakenet setup.

dr-frmr:
Reviewing and deploying on prerelease testnet to make sure we don't break anything

app/sequencer.hoon
dr-frmr: I think we can safely delete this whole comment block

hosted-fornet: 0acbc0cc removes comment block

dr-frmr (approved):

PR #303, opened 2022-11-29
Wallet: provide responses to apps that build transactions and send through us

dr-frmr:
The purpose of this PR is to add a scheme to the wallet agent where apps can be notified of completed transactions that they initiated. I'm actively using it in %pokur where this behavior is very necessary.

I considered routing these pokes through %uqbar but it actually doesn't make a difference because the receiving app does not yet have a way in gall to distinguish the "source app". So it just doesn't matter and I'll keep them going straight from %wallet. If this changes in the future will app sandboxing we can update then.

This is a breaking change to %wallet agent state so it's going to `next/smart`

### DIFF
--- a/app/wallet.hoon
+++ b/app/wallet.hoon
@@ -447,18 +447,23 @@
       ~&  >>>  "%wallet: couldn't find transaction hash for update!"
       $(unfinished t.unfinished, still-looking [i.unfinished still-looking])
     ::  put latest version of tx into transaction-store
-    =/  updated
+    =/  updated=[@ux =origin transaction:smart supported-actions output:eng]
       =+  found=(~(got by transactions.tx-latest) hash.i.unfinished)
       ::  add 200 to finished status code to get wallet status equivalent
       =.  status.transaction.found  (add 200 status.transaction.found)
-      [hash.i.unfinished origin.i.unfinished transaction.found action.i.unfinished output.found]
+      :*  hash.i.unfinished
+          origin.i.unfinished
+          transaction.found
+          action.i.unfinished
+          output.found
+      ==
     ::  when we have a finished transaction, use transaction origin to
     ::  notify an app about their completed transaction.
     %=  $
       unfinished  t.unfinished
         cards
       :-  (finished-tx-update-card updated)
-      ?~  -.+.updated  cards
+      ?~  origin.updated  cards
       [(notify-origin-card our.bowl updated) cards]
         transaction-store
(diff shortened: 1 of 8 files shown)


dr-frmr:
Another thing I'm considering for this PR is how the wallet agent might manage automatic signing and submitting of certain transactions. The use case for this is being able to have certain apps' transactions go straight through without manual signing. This is just a UX thing for some apps, but for others, it's necessary to automate behaviors.

For example, the %pokur app has a host controlling the payouts from a sit-n-go tournament. At the end of the tournament, the host ship should automatically calculate the payouts for winner(s) and send it to them. No one should have to go in and manually sign those transactions -- a single host should be able to passively manage many concurrent games.

app/wallet.hoon
```diff
@@ -418,24 +441,28 @@
...
-      [hash.i.unfinished transaction.found action.i.unfinished output.found]
+      [hash.i.unfinished origin.i.unfinished transaction.found action.i.unfinished output.found]
+    ::  when we have a finished transaction, use transaction origin to
+    ::  notify an app about their completed transaction.
     %=  $
       unfinished  t.unfinished
-      cards       [(finished-tx-update-card updated) cards]
+        cards
+      :-  (finished-tx-update-card updated)
+      ?~  -.+.updated  cards
```
hosted-fornet, line 461: Would be nice to give `origin` face to `origin.i.unfinished` above in 454 and refer to it as such here to avoid `-.+....`

lib/zig/wallet.hoon
```diff
@@ -16,14 +16,25 @@
   [%give %fact ~[/tx-updates] %wallet-frontend-update !>(-)]
 ::
 ++  finished-tx-update-card
-  |=  in=[@ux transaction:smart supported-actions output:eng]
+  |=  in=[@ux origin transaction:smart supported-actions output:eng]
   ^-  card
-  =+  `wallet-frontend-update`[%finished-tx in]
+  =+  `wallet-frontend-update`[%finished-tx [-.in +.+.in]]
```
hosted-fornet, line 21: This is ugly.

hosted-fornet (approved):
Changes look pretty straightforward; made a couple style comments, otherwise LGTM.

Once app sandboxing is real we should revisit this and think about how to avoid bad people taking advantage of this auto-submit mechanism, but for now I buy the argument that security is a larp, so there isn't much we can do.

dr-frmr:
Merging with nits fixed

Issue #304, opened 2022-11-29
indexer: crashed scry for bad transaction


bacwyls:
`0xa4ad.26bf.4635.6a0b.9fcb.feb4.3f3b.1f51` is a transaction I sent which crashed in contract execution. for context, it was a malformed zig send. I'm pretty sure it crashed because I sent with the item=@ux from uqbar-core README instead of hashing my account id.

When looking at this transaction using Explorer, the web page is an infinite spinner due to a `peek bad result`

## reproduce:
try to GET
http://myurbit.com/~/scry/indexer/json/transaction/0xa4ad.26bf.4635.6a0b.9fcb.feb4.3f3b.1f51.json

It should 404 and create a stack trace in dojo
```
peek bad result
/lib/zig/jold/hoon:<[63 34].[63 40]>
/lib/zig/jold/hoon:<[63 18].[63 41]>
/lib/zig/jold/hoon:<[68 40].[68 45]>
/lib/zig/jold/hoon:<[68 17].[68 46]>
/lib/zig/jold/hoon:<[68 16].[68 52]>
/lib/zig/jold/hoon:<[65 3].[69 5]>
/lib/zig/jold/hoon:<[64 3].[69 5]>
/lib/zig/jold/hoon:<[63 3].[69 5]>
/lib/zig/jold/hoon:<[62 3].[69 5]>
/lib/zig/jold/hoon:<[61 3].[69 5]>
/lib/zig/jold/hoon:<[60 3].[69 5]>
/lib/zig/jold/hoon:<[65 3].[69 5]>
/lib/zig/jold/hoon:<[64 3].[69 5]>
/lib/zig/jold/hoon:<[63 3].[69 5]>
/lib/zig/jold/hoon:<[62 3].[69 5]>
/lib/zig/jold/hoon:<[61 3].[69 5]>
/lib/zig/jold/hoon:<[60 3].[69 5]>
/lib/zig/jold/hoon:<[65 3].[69 5]>
/lib/zig/jold/hoon:<[64 3].[69 5]>
/lib/zig/jold/hoon:<[63 3].[69 5]>
/lib/zig/jold/hoon:<[62 3].[69 5]>
/lib/zig/jold/hoon:<[61 3].[69 5]>
/lib/zig/jold/hoon:<[60 3].[69 5]>
/lib/zig/jold/hoon:<[59 3].[69 5]>
/lib/zig/jold/hoon:<[58 3].[69 5]>
[...]

PR #305, opened 2022-11-29
indexer: virtualize jold to avoid crashing on malformed data

hosted-fornet:
Addresses #304 by virtualizing the `jold` call so that malformed data does not crash, and instead gives the default "un-jolded" output.


dr-frmr (approved):
LGTM, glad our friend caught this!

PR #308, opened 2022-11-30
smart: overhaul contract output typing

dr-frmr:
Long term project: overhaul and unify the way we handle untyped nouns that contracts output.

There are 3 important nouns that outside programs need to handle from contracts:
- transaction calldata: `(pair @tas *)`
- events: `(pair @tas *)`
- data items: `[label=@tas noun=*]`

Currently we have types and interface jolds posted on-chain (optionally). These will be replaced by a path where the final value is the hash of a core. That core will be similar to how `mar` files work for gall apps today, but it will differ slightly -- rather than covering a single `mark`, it should handle all the `@tas` labels a contract produces across transactions/events/data.

Instead of `grab`/`grow` as the main arms of our `mar` core, we will have `data`, `action`, and `event`. Each of these will "grow" an on-chain raw noun to a typed noun. Developers can provide `noun` and `json` arms within each of these main arms.

When a contract is deployed, the developer can optionally add an `interface` pith (typed path) to the `pact`. This can be a scry path that serves as a place to find the actual conversion core, but doesn't have to be. No matter what, though, the final value in the path must be the hash of the core.

Indexers will want to use these mark-interpreting cores just like they used to use the json mappings in `interface`/`types`. We will build a component that stores these cores in the indexer alongside contracts, such that indexers can make json representations to display in the block explorer, and also distribute the cores themselves to users.


dr-frmr:
This is "finished": I've removed old contract interfaces and added new ones for our main system-relevant contracts. The system works as desired, but future work will be needed for indexer / block explorer frontends to actually use these things. That will be incremental work that won't require breaking changes, though, so we can do all that later.

PR #310, opened 2022-12-02
Engine bugfix: don't enforce budget on zigs transactions that are part of continuation calls

dr-frmr:
This came up when performing a zigs %give from a contract inside a continuation call. Since the initial transaction caller pays the budget for all continuation calls, the budget should not be a part of the zigs token transfer validation in the contract. To nullify that line of logic in the contract, we can just set the field to 0 in the engine.

In the future I want to remove the budget field from zigs contract entirely if possible, and do all this logic in the engine.

### DIFF
--- a/lib/zig/sys/engine.hoon
+++ b/lib/zig/sys/engine.hoon
@@ -79,6 +79,7 @@
         ~&  >>>  "engine: tx failed gas audit"
         (exhaust bud.gas.tx %3 ~)
       ::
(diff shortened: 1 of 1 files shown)


dr-frmr:
This is good for `master` -- I'll put my more extensive solution in `next/smart` since it involves changes to the `zigs.hoon` contract.

hosted-fornet (approved):
I see what is going on here and the solution LGTM :+1:

PR #314, opened 2022-12-02
Da/subscribe

tadad:


tadad:
A few things in this PR
- basic subscription testing working
  - This is just a simple first draft that let's you verify that a subscription was received by an agent on another ship
- caching of ship snapshots
  - doesn't work fully because of a zuse mismatch, but the problem gets fixed in [this](https://github.com/urbit/urbit/pull/6044) PR, so it'll work soon.
- updates to the README

app/subscriber.hoon
```diff
@@ -29,65 +29,44 @@
...
-    =/  agent  i.t.t.path
-    =/  pafth  t.t.t.path
-    =/  qew  (~(got bi facts.state) [ship agent] pafth)
-    =/  fakts  ~(tap to qew)
-    ``noun+!>(fakts)
+      [%x %signs @ @ ^]
+    =/  =ship  (slav %p i.t.t.path)
+    =*  app  i.t.t.t.path
+    =*  paf  t.t.t.t.path
+    ``noun+!>((~(got by signs.state) [ship app paf]))
```
hosted-fornet, line 69: It might be nice to change this to return the `get by` to avoid crashing? E.g. try

```
=my-test-2 [[%read-subscription [~bud ~nec %rollup /not-a-sub-path/noun] '']]~
:ziggurat &ziggurat-action [%foo %add-and-queue-test `%my-test-2 ~[/zig/sur/zig/indexer/hoon] my-test-2]
:ziggurat &ziggurat-action [%$ %run-queue ~]
```

ted/ziggurat/test/run.hoon
```diff
@@ -83,6 +83,45 @@
...
+          /(scot %p who.payload)/subscriber/[now]
+          /facts/(scot %p to.payload)/[app.payload]
+          path.payload
+          /noun
+        ==
+    ==
+  ?.  ?=(^ scry-noun)  (pure:m !>(~))
+  =+  fact-set=(need ;;((unit (set @t)) scry-noun))
+  ?:  (~(has in fact-set) expected)
+    (pure:m !>(~))
```
hosted-fornet, line 105: I would recommend not using `~` to indicate success:

1. It interferes with line 102 above, where if we get a `~` result from scry we are informing caller that our scry was successful: this is wrong.
2. It means we place a `~` in the `result` field of a `test-result`. We should always have `expected.test-result == result.test-result` if `success=%.y`. The reason is that we may (or users may) make use of the `result.test-result`, and so it should be a predictable value when `success=%.y`, and not vary its semantics based on which test step type has produced it.

ted/ziggurat/test/run.hoon
```diff
@@ -233,9 +272,37 @@
...
   ::
-      %read-subscription  !!
+        %read-subscription
+    ::  TODO: not sure if this should actually be a vase
+    ::  when res-text is '~' that indicates success. Otherwise
+    ::  the rest of the set is passed back to show you what other
+    ::  subscriptions were received
+    ;<  result=vase  bind:m
+      (read-pyro-subscription [payload expected]:test-step)
+    ::  transform to qeu, cut it at the
```
hosted-fornet, line 282: Remove old comment.

ted/ziggurat/test/run.hoon
```diff
@@ -233,9 +272,37 @@
...
+        %read-subscription
+    ::  TODO: not sure if this should actually be a vase
+    ::  when res-text is '~' that indicates success. Otherwise
+    ::  the rest of the set is passed back to show you what other
+    ::  subscriptions were received
+    ;<  result=vase  bind:m
+      (read-pyro-subscription [payload expected]:test-step)
+    ::  transform to qeu, cut it at the 
+    =*  expected  expected.test-step
+    =/  res-text=@t  (crip (noah result)) :: TODO probably wrong
```
hosted-fornet, line 284: Would be nice to `tap by` the `set` when we fail to find our `expected` (perhaps that should happen above near line 100? Not sure). The reason is that it will be simpler to parse for us or for FEs: `set`s get displayed in the `n l r` form.

ted/ziggurat/test/run.hoon
```diff
@@ -233,9 +272,37 @@
...
-      %subscribe  !!
+      %subscribe
+    ;<  ~  bind:m  (send-pyro-subscription payload.test-step)
+    ;<  trs=test-results:zig  bind:m
+      %:  run-steps
+          project-id  test-id
+          `test-steps:zig`expected.test-step  ~
+      ==
+    ?~  tr=(test-results-of-reads-to-test-result trs)
+      ~|("ziggurat-test-run: %poke expected can only contain %scrys, %subscribes, %waits" !!)
```
hosted-fornet, line 300: Change text since we are in `%subscribe` here

ted/ziggurat/test/run.hoon
```diff
@@ -83,6 +83,45 @@
...
+        ;:  weld
+          /(scot %p our.bowl)/pyro/[now]/i/(scot %p who.payload)/gx
+          /(scot %p who.payload)/subscriber/[now]
+          /facts/(scot %p to.payload)/[app.payload]
+          path.payload
+          /noun
+        ==
+    ==
+  ?.  ?=(^ scry-noun)  (pure:m !>(~))
+  =+  fact-set=(need ;;((unit (set @t)) scry-noun))
```
hosted-fornet, line 103: We should probably not use `+need` here, since it crashes on fail. Instead can go a `?~` or something. If we remove `~` return meaning success as suggested below on line 105, could do something like

`?~  fact-set  (pure:m !>(~))`

The reason we should avoid crashes if possible is so that even if one test-step fails, we can still attempt to run others.

hosted-fornet (changes requested):

app/subscriber.hoon
tadad: changed it to this because I don't want to unwrap a unit
```
    :^  ~  ~  %noun
    =+  (~(get by signs) [ship app paf])
    ?@  -  !>(~)  !>((need -))
```

app/subscriber.hoon
```diff
@@ -29,65 +29,48 @@
...
-    =/  qew  (~(got bi facts.state) [ship agent] pafth)
-    =/  fakts  ~(tap to qew)
-    ``noun+!>(fakts)
+      [%x %signs @ @ ^]
+    =/  =ship  (slav %p i.t.t.path)
+    =*  app  i.t.t.t.path
+    =*  paf  t.t.t.t.path
+    :^  ~  ~  %noun
+    =+  (~(get by signs) [ship app paf])
+    ?@  -  !>(~)  !>((need -))
```
hosted-fornet, line 73: nit: Slight preference to use `?~` or something else here that denotes we are checking if it is `~`, rather than `@`: better to be explicit than implicit.

app/ziggurat.hoon
```diff
@@ -674,6 +683,16 @@
         !>([%$ %run-queue ~])
       ~
     ==
+  ::
+      ::  XX the /effect/restore is used by some other vanes in pyro
+      ::     so we might want to edit that in pyro so that ship-ready
+      ::     effects are a different subscribe path
```
hosted-fornet, line 689: I'm not really sure what this comment is about. Do you want to discuss/address this on this PR or on a subsequent PR when we get snapshots working more robustly?

ted/ziggurat/test/run.hoon
```diff
@@ -83,6 +83,45 @@
...
+          /facts/(scot %p to.payload)/[app.payload]
+          path.payload
+          /noun
+        ==
+    ==
+  ?.  ?=(^ scry-noun)  (pure:m !>(~))
+  =+  fact-set=(fall ;;((unit (set @t)) scry-noun) *(set @t))
+  ?:  (~(has in fact-set) expected)
+    (pure:m !>(expected))
+  (pure:m !>((crip (noah !>(~(tap in fact-set))))))
```
hosted-fornet, line 106: We may need to come back to this, but it is good for now. My concern looking at the output is that we can't `+ream` it because it is in the `wain` form.

hosted-fornet (approved):
Some comments on readme and a few nits, otherwise LGTM, nice work :+1: 

PS: oh, also please update the pill before merging.

readme-pyro.md
```diff
@@ -186,54 +149,59 @@ An alternative to `send-nec`:
...
+
+:pyro|init ~bud
+:pyro|dojo ~bud ":indexer &set-sequencer [~nec %sequencer]"
+:pyro|dojo ~bud ":indexer &set-rollup [~nec %rollup]"
+:pyro|dojo ~bud ":indexer &indexer-bootstrap [~nec %indexer]"
+:pyro|dojo ~nec ":uqbar &wallet-poke [%import-seed 'post fitness extend exit crack question answer fruit donkey quality emotion draw section width emotion leg settle bulb zero learn solution dutch target kidney' 'squid' 'nickname]"
+
+:pyro &action [%snap-ships /testnet ~[~nec ~bud]]
+:pyro &action [%export-snap /testnet]
+|unmount %zig
```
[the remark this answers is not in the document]
tadad: is there another way to push a desk down to unix?

app/ziggurat.hoon
tadad: no this comment is wrong - I think I misunderstood /restore whenever I wrote it

PR #317, opened 2022-12-05
Da/commit

tadad:
add `:pyro|commit` to change files


hosted-fornet:
Can you give some more details on usage? I tried:
```hoon
:pyro &pill +zig!solid %base
:ziggurat &ziggurat-action [%foo %start-pyro-ships ~]
:pyro|commit ~nec %zig
```

and got
```
[%bad-mount-point-from-unix %zig]
```

so tried
```hoon
:pyro|dojo ~nec "|merge %zig our %base"
:pyro|commit ~nec %zig
```

with same result.

tadad:
@hosted-fornet right forgot to mention that - you do have to run `:pyro|dojo ~nec "|mount %zig"` before using commit

hosted-fornet (commented):
One thing that would be very desirable would be if we could delete files in a desk if not passing `files` argument. As is, the desk will have all its existing files that are not overwritten unchanged; really we want a copy of the desk from the fakeship to the virtualship.

I could be convinced that the deleting files thing should be another PR. How are you imagining this tool (and others if they are necessary) will be integrated with %ziggurat for initial setup of a ship and subsequent syncing of a ship during development from FE?

hosted-fornet:
Some more specific steps for integration with %ziggurat, for here or another PR:

1. `+on-init`; distribute `%base` pill and copy in %zig rather than distributing `%base %zig` pill.
2. On project creation, copy that new desk into the ships (if they are running; else, when start, copy desk in (?))
3. Integrate so that when we save a file to a project it is also copied to the virtualships
4. Make sure we are snapshotting in a way that works with these systems

hosted-fornet (approved):
Thinking more: it makes the most sense to keep this PR tightly scoped. We can add those other features I mention above in another PR. Take a look at the comments and then merge when ready :+1:

tadad:
> Some more specific steps for integration with %ziggurat, for here or another PR:
> 
> 1. `+on-init`; distribute `%base` pill and copy in %zig rather than distributing `%base %zig` pill.
> 2. On project creation, copy that new desk into the ships (if they are running; else, when start, copy desk in (?))
> 3. Integrate so that when we save a file to a project it is also copied to the virtualships
> 4. Make sure we are snapshotting in a way that works with these systems

Agreed

PR #319, opened 2022-12-05
sur/engine: remove unnecessary import

dr-frmr:
Inspired to do this because it drastically reduces the number of files other desks need to copy in to interact with wallet and indexer!


hosted-fornet (approved):
LGTM. Is this or some other similar dep expected to come back in later? Or was this just a leftover?

PR #320, opened 2022-12-06
show default types in comments in gall, fix missing fas

0x70b1a5:


lib/zig/ziggurat.hoon
```diff
@@ -686,20 +686,22 @@
   +*  this     .
       default   ~(. (default-agent this %|) bowl)
   ::
-  ++  on-init
+  ++  on-init                     :: `agent
```
hosted-fornet, line 689: Should this and below be `[(list card) agent]` or `[(list card) this]` or something to that effect, rather than tic-agent? Gives more useful info, but not sure if goal here is to show default type or more general return type.

0x70b1a5: I like that idea, and I trust your opinion as a hoon dev

lib/zig/ziggurat.hoon
```diff
@@ -686,20 +686,23 @@
...
```
hosted-fornet, line 689: Missed the one `+on-init`

hosted-fornet (approved):
You missed the type update on `+on-init`, but otherwise LGTM :+1:

PR #323, opened 2022-12-07
Da/iris

tadad:
Iris requests in pyro are working


hosted-fornet (approved):
Some style comments: otherwise looks good and works well. Also, please update the pill before merging. I've created #325 so we can come back to implement the remainder of %iris functionality later.

PR #324, opened 2022-12-08
ziggurat: watch clay

hosted-fornet:
We were previously only doing project updates and contract compile checks on %save-file and the like. This is one way to add files (from IDE), but another is to edit on unix and `|commit`, so we need to watch clay and send project updates/contract compiles based on that.

Also bugfixes and dev QOL changes.


lib/zig/ziggurat.hoon
```diff
@@ -782,7 +813,8 @@
 ++  get-state-to-json
   |=  [p=project our=@p now=@da]
   ^-  json
-  (state-to-json (get-state p our now))
+  ?~  state=(get-state p our now)  ~
+  (state-to-json state)
```
hosted-fornet, line 817: This is for when `+new-project` calls `+make-read-desk`, which itself calls `+make-project-update` which calls this gate.. `+get-state` does a scry out to `%pyro` to get chain info, but chain has not been started yet; have modified it now so that scrying with no setup piers on %pyro gives a `~`.

hosted-fornet:
@dr-frmr sorry, no review req.d; I misclicked you instead of @tadad

PR #326, opened 2022-12-09
ziggurat: move `test-steps`, `custom-step-definitions` to files

hosted-fornet:
Move definitions of `test-steps` and `custom-step-definitions` to files. This allows these to be checked in to a repo. It also makes them accessible to outside text editors, further enabling the "BYOTextEditor" work flow we anticipate many users will prefer.

I noticed while testing this PR that adding `test-steps` is pretty slow and traced it to the `%ca` scry in `compile-test-surs`. We can speed this up by caching these imports; I'll add that in another PR: I want to split out the state into core/inflated.


app/ziggurat.hoon
```diff
@@ -84,115 +85,98 @@
...
     =.  custom-step-definitions.test
       %+  ~(put by custom-step-definitions.test)  tag
-      [custom-step-definition compilation-result]
-    test
+      [p compilation-result]
+    ~?  ?=(%| -.compilation-result)
+      %ziggurat^%custom-step-compilation-failed^p.compilation-result
+    `test
   ::
   ++  add-test
```
tadad, line 133: is there any way to speed up compilation?

zig/test-steps/send-nec-addresses.hoon
```diff
@@ -0,0 +1,23 @@
...
+        :^  ~nec  %uqbar  %wallet-poke
+        '[%submit from=(~(got by addresses) ~nec) hash=0xa99c.4c8e.1c8d.abb8.e870.81e8.8c96.2cf5 gas=[rate=1 bud=1.000.000]]'
+      ~
+    :+  %dojo  [~nec ':sequencer|batch']
+    :_  ~
+    :+  %scry
+      :-  ~nec
+      :^  'update:indexer'  %gx  %indexer
+      /newest/item/0x89a0.89d8.dddf.d13a.418c.0d93.d4b4.e7c7.637a.d56c.96c0.7f91.3a14.8174.c7a7.71e6/noun
+    ''
```
tadad, line 21: talked to ted about how to get expected to work here such that certain values are ignored - basically what you want is a nest-check (can probably just use `+soft`), then a list of wing-name and value. Use `%=` to "block out" any value that we don't care about. Should probably write this up separately somewhere

readme-pyro.md
```diff
@@ -98,47 +98,62 @@ To avoid the overly-verbose scries, you can also use the `+scry` generator which
...
+:ziggurat &ziggurat-action [%foo %start-pyro-ships ~[~nec ~bud]]
+:ziggurat &ziggurat-action [%$ %run-queue ~]
+
+::  Tell `%ziggurat` not to run any more tests right now.
+::   Also resets state when `%start-pyro-ships` is called again.
 :ziggurat &ziggurat-action [%foo %stop-pyro-ships ~]
-:ziggurat &ziggurat-action [%foo %start-pyro-ships ~]
 ```
 
 An alternative to `send-nec`:
```
tadad, line 159: should just delete this section

tadad (approved):
just a few quickquestions and style comments. Added one commit because lib/ziggurat wasn't compiling

zig/test-steps/send-nec-addresses.hoon
hosted-fornet: Cool. Ok, yeah I'd want to see an example for a simple case or two as well, so please include that in the writeup. One highly desirable feature of this system is it should be easy and intuitive to use; ideally a new Hooner should be able to pick it up without difficulty. And it should also ideally be very readable.

app/ziggurat.hoon
hosted-fornet: I agree it is noticeably slow. We can cache the %ca scries used to bring in libraries when building `test-steps` subjets, which is the slowest part. At least based on what I was seeing during testing, that was the lionshare of compilation time, so for users who import the same libraries frequently, we'll speed things up a lot.

readme-pyro.md
hosted-fornet: Once we've added a few more PRs let's do a more thorough re-consideration and re-write of the readme. This section is probably not that useful, I agree. I was imagining it might be useful to demonstrate more complex series of `:pyro|dojo` calls to show that it really does just work like the Dojo people know and love.

hosted-fornet:
> just a few quickquestions and style comments. Added one commit because lib/ziggurat wasn't compiling

Thanks for the fix.

PR #327, opened 2022-12-12
Da/subscriber start

tadad:
- Quick bug fix in pyre's on-init
- Got rid of `%subscriber` from desk.bill

later if we get rid of the `%zig` desk from the pill we will probably have to change this


hosted-fornet:
LGTM, please merge in next/pyro and update pill to guarantee it is most recent before merging :+1:

PR #328, opened 2022-12-12
Da/commit plus

tadad:


hosted-fornet:
I like the simplification of the logic, but what was your reasoning for removing %insert-files poke from %pyro?

tadad:
> I like the simplification of the logic, but what was your reasoning for removing %insert-files poke from %pyro?

I don't think anything relies on it, and I think it's cleaner to just pass in raw events via generators than it is to bloat pyro's logic with a bunch of stuff relating to specific arvo tasks

hosted-fornet:
The reason I like having the %insert-files poke is that it exposes a nicer interface for other people to build on. There are tiers of knowledge about urbit, and the majority of urbit devs don't really need to know about unix events/effects -- e.g. I didn't know anything about events/effects until working on this project. People fresh out of app school will understand pokes, but if they see `%event who /c/sync/0v1n.2m9vh %info desk %&` that is a whole other thing they need to understand before they can build on it. At best, they can copy-paste it from our code, but this stuff is not well documented. I think it makes sense to hide this complexity where we can to make it easy for people to build on and understand at a high level what is going on in this code if they want to.

tadad:
people can still build on it - they just have to use the `:pyro|commit` generator, which wraps up the logic for them nicely.

tadad:
:pyro|commit is a little slow (5-10secs which isn't terrible). It has something to do with ford caches so I'm going to take a look at that soon

PR #329, opened 2022-12-13
zigs.hoon: add %push and %pull from fungible standard

dr-frmr:


dr-frmr:
Effectively using this on zigs contract in %pokur's escrow contract. See here: https://github.com/dr-frmr/pokur/blob/39a2100ea8ca94c35f38021a926729e691c05873/src/con/escrow.hoon#L173

tadad (approved):

PR #331, opened 2022-12-14
ziggurat: get %deploy-contract working

hosted-fornet:
requires loom be 4GB, so must use 1.13 or newer
binaries and start ship using `--loom 32`, e.g.:
```bash
urbit --loom 32 nec
```

TODOs:
* Speed things up (next up, via caching so we don't do so many `%ca`s)
* Track down the memory blowup. You can experience this yourself by running your ship with 2GB loom. For some reason the `:pyro|dojo` paste of the contract deployment (which is many thousands of characters) crashes with OOM. This is weird because it does not crash when run by the normal Dojo; only the Pyro Dojo.

Also includes a variety of unrelated bugfixes and improvements; I'll annotate them.


app/ziggurat.hoon
```diff
@@ -186,22 +183,23 @@
...
+            p  [%face %test-globals p.initial-test-globals]
+        ==
       %+  roll  surs
       |:  [[face=`@tas`%$ sur=`path`/] subject=`^vase`!>(..zuse)]
-      ?:  %-  %~  has  in
-              %-  %~  gas  in  *(set @tas)
-              ~[%addresses %test-results]
-          face
-        ~|("%ziggurat: compilation failed; cannot use %addresses, %test-results: reserved and built into subject already" !!)
+      ?:  =(%test-globals face)
```
hosted-fornet, line 201: Now put test globals inside `test-globals` rather than just a variety of unrelated faces.

app/ziggurat.hoon
```diff
@@ -320,26 +319,103 @@
...
+                ~
+            ::
+                :+  %poke
+                  :^  ~nec  %uqbar  %wallet-poke
+                  %-  crip
+                  "[%submit from={<address>} hash=deploy-tx gas=[rate=1 bud=1.000.000]]"
+                ~
+            ::
+                [%dojo [~nec ':sequencer|batch'] ~]
+            ==
```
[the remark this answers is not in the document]
hosted-fornet: This pattern makes the need for `test-results:test-globals` accessible by subsequent test steps less immediate. I think that can still be really nice, and is something I want to add, but it is no longer something that we have no workaround for.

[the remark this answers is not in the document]
hosted-fornet: One thing I was considering here was adding an additional `test-step` type: `%dojos`. Something like a list of `%dojo`s to be used in cases like the above. It might be slightly nicer for users, but it is not necessary, so I did not add it for now.

app/ziggurat.hoon
```diff
@@ -221,6 +219,7 @@
   ++  handle-poke
     |=  act=action
     ^-  (quip card _state)
+    ?>  =(our.bowl src.bowl)
```
hosted-fornet, line 222: Don't let other people hack my %ziggurat.

lib/py/io.hoon
```diff
@@ -106,7 +106,7 @@
   |=  [=ship =tape]
   =/  m  (strand ,~)
   ^-  form:m
-  ~&  >  "dojo: {tape}"
+  :: ~&  >  "dojo: {tape}"
```
hosted-fornet, line 109: Was spammy and we already get this info from pyro's dill stuff. Somewhat useful for debugging when a command to dojo doesn't go through.

tadad: I would consider just getting rid of these files. We haven't changed them from the ph/ thread files and we aren't using these anymore

tadad (approved):

tadad:
@hosted-fornet just wrote #335 so we may want to merge that before commiting this so that we don't have to use 4GB loom, use dojo for all I/O

lib/py/io.hoon
hosted-fornet: Good point, making a new PR to change this now.

hosted-fornet:
> @hosted-fornet just wrote #335 so we may want to merge that before commiting this so that we don't have to use 4GB loom, use dojo for all I/O
> 
> for the `%poke` test-step we can use the `%poke` that I write there, and for `%custom-write` we can use the `%task` action I wrote for pyro - which allows for any arbitrary input, typed, into a pyro ship

Okay, nice. I'll have to take a look at this PR. My preference is to merge this here since it is working and then we can write a new PR once we've got #335 merged so we don't have to use 4GB loom anymore.

app/ziggurat.hoon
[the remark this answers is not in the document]
hosted-fornet: Yeah, I agree we should stop using Dojo as main input method, but it is still a valuable tool for development, testing, etc.

PR #332, opened 2022-12-14
ziggurat: cache %ca scries

hosted-fornet:
WIP


hosted-fornet:
To test try `%add-and-queue`ing %scry-nec and then %scry-bud. The first takes ~s5 on my system; but once the imports have been cached, the second is effectively instant.

tadad (approved):

PR #334, opened 2022-12-15
ziggurat: re-add `%add-test` with `test-steps`

hosted-fornet:
for use with frontend.

expectation is more developed tests can be imported from files using what has been retitled `%add-test-file`.
these can also be editing using the standard text editing UI or an external text editor.

during development, can have a pane where user can write in `test-steps` for more integrated initial development

Still TODO: add another poke to `%save-test` to convert a test into a file and save it in the project dir.

### DIFF
--- a/app/ziggurat.hoon
+++ b/app/ziggurat.hoon
@@ -129,6 +129,59 @@
     `test
   ::
   ++  add-test
+    |=  $:  project-name=@t
+            name=(unit @t)
+            surs=test-surs
+            =test-steps
+        ==
+    ^-  [[(list card) test] _state]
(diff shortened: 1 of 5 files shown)


hosted-fornet:
FYI @0x70b1a5 in particular see

https://github.com/uqbar-dao/uqbar-core/blob/ee286ed1e010af31d0552722b2d7ae178f6d0bfb/sur/zig/ziggurat.hoon#L122-L125

https://github.com/uqbar-dao/uqbar-core/blob/ee286ed1e010af31d0552722b2d7ae178f6d0bfb/readme-pyro.md#alternative-test-steps-input

PR #336, opened 2022-12-15
ziggurat: clean up py ph and rearrange

hosted-fornet:


hosted-fornet:
As per @tadad suggestion: https://github.com/uqbar-dao/uqbar-core/pull/331#discussion_r1049815768

Also moved some stuff around: the remaining contents in lib/py -> zig/snapshots

app/pyro.hoon
```diff
@@ -66,7 +66,6 @@
     :_  this
     :~  [%pass / %agent [our dap]:bowl %poke %pill !>(cached-pill)]
         [%pass / %agent [our dap]:bowl %poke %action !>([%import-snap /testnet/jam /testnet])]
-      :: [%pass / %agent [our dap]:bowl %poke %action !>([%import-fresh-piers /zig/lib/py/fresh-piers/jam])]
```
tadad, line 69: let's keep this because I want to uncomment this as soon as the ford-cache PR is merged into core

lib/zig/pyro.hoon
```diff
@@ -0,0 +1,40 @@
...
+  |=  [who=ship what=tape]
+  ^-  (list aqua-event)
+  %+  send-events-to  who
+  ^-  (list unix-event)
+  :~
+    [/d/term/1 %belt %ctl `@c`%e]
+    [/d/term/1 %belt %ctl `@c`%u]
+    [/d/term/1 %belt %txt ((list @c) what)]
+    [/d/term/1 %belt %ret ~]
+  ==
```
tadad, line 26: the `%dojo` poke is this exact code so we could delete this file and just send a raw poke directly to pyro

app/pyro.hoon
hosted-fornet: Cool. We'll need to do some add'l work on this import/export when we bring it back in, but we can worry about that later.

lib/zig/pyro.hoon
hosted-fornet: Okay, let's get your poke PR merged, then come back to this PR. There is also some work that needs to be done between your poke PR and then, because we need to get the pyro part interfaced with the ziggurat run-tests part.

tadad: I'm not even talking about the code I just wrote - I'm talking about this in app/pyro.hoon
```
  ?-    -.act
      %dojo
    :_  state
    %-  send-events
    ^-  (list aqua-event)
    %+  turn
      ^-  (list unix-event)
      :~  [/d/term/1 %belt %ctl `@c`%e]
          [/d/term/1 %belt %ctl `@c`%u]
          [/d/term/1 %belt %txt ((list @c) ;;(tape command.act))]
          [/d/term/1 %belt %ret ~]
      ==
    |=  ue=unix-event
    [%event who.act ue]
  ::
```

hosted-fornet: Ah, I misunderstood. I think I must've missed this %pyro interface change; it never got used in %ziggurat. We can potentially change ted/ziggurat/test/run to make use of this %pyro poke.

tadad: I'm also okay with keeping the pyro-inputs super simple (just aqua-events/unix-events), and pushing any specific inputs (dojo commands, tasks/notes/etc.) out to generators or threads. Would make pyro much cleaner

PR #337, opened 2022-12-15
ziggurat: integrate %pyro poke primitive

tadad:


app/ziggurat.hoon
```diff
@@ -376,35 +377,45 @@
...
+                  =/  new-pending=(set @ux)
+                    %~  key  by
+                    !<((map @ux *) result:i:new-test-result)
+                  =/  diff-pending=(list @ux)
+                    ~(tap in (~(dif in new-pending) old-pending))
+                  ?>  ?=([@ ~] diff-pending)
+                  :^  %submit  from={<address>}
+                    hash=i:diff-pending
+                  gas=[rate=1 bud=1.000.000]
+                  """
```
hosted-fornet, line 418: The pattern here in %deploy-contract for submitting to %wallet is something I'll break out into a custom-step-definition in a new PR Soon. The pattern is:

1. Scry %wallet `/pending-store` for current state
2. Send tx
3. Scry %wallet `/pending-store` for new state
4. Diff the `/pending-store` scries -> get our pending tx hash
5. Sign that tx
6. Send a new batch to %sequencer

lib/zig/ziggurat.hoon
```diff
@@ -267,6 +267,21 @@
...
+++  show-test-result
+  |=  =test-result
+  ^-  shown-test-result
+  %+  turn  test-result
+  |=  [success=? expected=@t result=vase]
+  =/  res-text=@t  (crip (noah result))
+  :+  success  expected
+  ?:  (lte 1.024 (met 3 res-text))  '<elided>'
+  res-text
+::
```
hosted-fornet, line 284: These support the change to `test-results` where `result` is now a `vase`. We display by `+noah`ing.

ted/ziggurat/test/poke.hoon
```diff
@@ -0,0 +1,29 @@
...
+  |=  args-vase=vase
+  ^-  form:m
+  =/  args  !<((unit arg-mold) args-vase)
+  ?~  args
+    ~&  >>>  "Usage:"
+    ~&  >>>  "-zig!ziggurat-test-poke who=@p to=@p app=@tas mark=@tas payload=@t"
+    (pure:m !>(~))
+  ;<  ~  bind:m  (poke:pyro [who to app mark payload]:u.args)
+  (pure:m !>(~))
+--
```
hosted-fornet, line 29: Just a helper since we need to use a ted to call `poke:pyro`.

ted/ziggurat/test/run.hoon
```diff
@@ -40,15 +41,26 @@
...
     (mule |.((slap-subject payload)))
-  ?:  ?=(%| -.compilation-result)
-    (trip payload)
-  (noah p.compilation-result)
+  ?:  ?=(%& -.compilation-result)  compilation-result
+  :-  %|
+  %-  crip
+  %+  roll  p.compilation-result
+  |=  [in=tank out=tape]
+  :(weld ~(ram re in) "\0a" out)
```
hosted-fornet, line 58: Use `+mule-slap-subject` to get more output when we fail to compile a `test-step` -- rather than just crashing.

hosted-fornet:
@tadad please review when you've got a chance.

My plan is to make a followup PR and there I'll cut down on the number of `send-nec` variants. I think it makes sense to keep a few around to demo some different ways to use `test-steps`, but this is getting out of hand!

app/ziggurat.hoon
hosted-fornet: Another thing that I want to work on is compilation error output from multi-line `test-step`s like this. This one was a pain to debug.

ted/ziggurat/test/run.hoon
```diff
@@ -186,8 +195,7 @@
     %-  slop  :_  subject
     results-vase(p [%face %test-results p.results-vase])
   |-
-  ;<  ~  bind:m
-    ?~(snapshot-ships (pure:(strand ,~) ~) (sleep ~s1))  :: TODO: unhardcode; tune time to allow previous step to continue processing
+  ;<  ~  bind:m  (sleep ~s1)  :: TODO: unhardcode; tune time to allow previous step to continue processing
```
hosted-fornet, line 198: Fixes a bug where the `expected` of write steps might not read out the proper value since the change had not propagated yet.

hosted-fornet: Hmm. Bug seems to still be here. This underscores the need for a non-hacky test-step discretization.

PR #342, opened 2022-12-20
interface JSON edits

dr-frmr:


con/lib/zigs-interface-types.hoon
```diff
@@ -52,19 +54,48 @@
...
+  ::
+  ++  push-cord
     ^-  cord
     '''
     [
       {"to": "ux"},
       {"amount": "ud"},
       {"from-account": "ux"},
-      {"to-account": ["unit", "ux"]}
+      {"calldata": "*"}
```
hosted-fornet, line 78: `calldata` here and in `fungible-interface-types` will fallback to `+noah`d output:

https://github.com/uqbar-dao/uqbar-core/blob/06325590e362db5d7e3d354a017ad78fae9a42e5/lib/zig/jold.hoon#L202-L213

hosted-fornet: Which is fine -- just noting it.

con/lib/fungible-interface-types.hoon
```diff
@@ -128,7 +126,6 @@
       {
         "mints": [
           {"to": "ux"},
-          {"account": ["unit", "ux"]},
           {"amount": "ud"}
         ]
       }
```
hosted-fornet, line 131: +mint-cord here looks off to me. I think it should read:

```hoon
  ++  mint-cord
    ^-  cord
    '''
    [
      {"token-metadata": "ux"},
      {
        "mints": [
          "list",
          [
            {"to": "ux"},
            {"amount": "ud"}
          ]
        ]
      }
    ]
    '''
```

hosted-fornet (approved):
One comment and one thing I think is a bug. Otherwise LGTM

con/lib/zigs-interface-types.hoon
dr-frmr: That's good! Should probably make this canonical with `%'*'` in the switch, even.

con/lib/fungible-interface-types.hoon
dr-frmr: Thanks, yeah, will fix

PR #346, opened 2022-12-21
ziggurat: make `custom-step-definitions` return `test-steps`

hosted-fornet:
also, make use of it to define `send-wallet-transaction.hoon`, a utility for sending writes to virtualnet.
it removes a lot of the boilerplate, so user/dev just inputs the %wallet-poke and the result is a `test-steps` that:
1. submits the %wallet-poke
2. signs the %wallet-poke
3. creates a new batch with the %sequencer

Working as of ec6f900. Still TODO:
1. ~~make the user/dev interface smoother. Ideally user/dev should not have to use `+noah`.~~ (not doing)
2. ~~can we remove custom-write vs custom-read? it is more confusing in this context~~ (not doing)
3. ~~clean up old test-steps definitions~~
4. ~~readme update~~


ted/ziggurat/test/run.hoon
```diff
@@ -1,8 +1,8 @@
-/-  spider,
+/-  pyro,
+    spider,
     zig=zig-ziggurat
 /+  strandio,
     pyro=zig-pyro,
-    test=zig-ziggurat-test,
     zig-lib=zig-ziggurat
```
tadad, line 6: do the two `pyro`s not cause conflicts?

tadad (approved):
looks good, everything ran fine for me

ted/ziggurat/test/run.hoon
hosted-fornet: Ah, good point, this is bad. I'll fix this.

PR #348, opened 2022-12-21
uqbar.hoon: add a poke to open faucets

dr-frmr:
This PR makes it so we can add a faucet UI to wallet. Since we track active sequencers in uqbar.hoon, it makes the most sense to attach this poke there -- sequencers will usually host faucets, and if this changes, uqbar.hoon should track faucet hosts. 

Tested on fakeships, will need to build usage into wallet FE.

### DIFF
--- a/app/uqbar.hoon
+++ b/app/uqbar.hoon
@@ -3,10 +3,10 @@
 ::  The agent for interacting with Uqbar. Provides read/write layer for userspace agents.
 ::
 /-  spider,
+    f=zig-faucet,
     u=zig-uqbar,
     ui=zig-indexer,
-    w=zig-wallet,
-    f=zig-faucet
+    w=zig-wallet
 /+  agentio,
     default-agent,
     dbug,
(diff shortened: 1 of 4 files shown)


app/uqbar.hoon
```diff
@@ -5,7 +5,8 @@
 /-  spider,
     u=zig-uqbar,
     ui=zig-indexer,
-    w=zig-wallet
+    w=zig-wallet,
+    f=zig-faucet
```
hosted-fornet, line 9: Tiny nit: would prefer the following order for sur/lib imports:
1. default-faced sur/libs, alphabetic
2. faced sur/libs, alphabetic by assigned face

so would involve changing the above to
```hoon
/-  spider,
    f=zig-faucet,
    u=zig-uqbar,
    ui=zig-indexer,
    w=zig-wallet
```

hosted-fornet (approved):
Added a nit, otherwise LGTM :+1:

app/uqbar.hoon
dr-frmr: sure, adding then merging

PR #356, opened 2022-12-29
pyro: rearrange pokes; allow FE pokes

hosted-fornet:
**Problem**:

Frontend cannot interact with %pyro.

**Solution**:

Rearrange %pyro and add `+enjs` and `+dejs` cores and appropriate `mar` files.

**Notes**:

~~WIP~~

I'll do a similar unification for %ziggurat scries/sub updates in #355 to reduce merge headaches.

### DIFF
--- a/app/pyro.hoon
+++ b/app/pyro.hoon
@@ -120,35 +120,55 @@
     |=  =path
     ^-  (unit (unit cage))
     ?+    path  ~
-        [%x %fleet-snap ^]  ``pyro-update+!>((~(has by fleet-snaps) t.t.path))
-        [%x %fleets ~]      ``pyro-update+!>(~(key by fleet-snaps))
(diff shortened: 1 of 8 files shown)


hosted-fornet:
FYI @0x70b1a5

app/pyro.hoon
```diff
@@ -721,14 +643,106 @@
   ::
       %clear-snaps
     `state(fleet-snaps ~)
+  ::
+      %pill
+    (poke-pill pill.act)
```
tadad, line 648: I think it's cleaner to just in line +poke-pill here

tadad (commented):
small style comment on poke-pill, otherwise looking good

app/pyro.hoon
hosted-fornet: Good call, inlined in 2980c95

PR #357, opened 2023-01-01
wallet: fixed ++  make-tokens bug

leowini:
**Problem**:

When you derive a new address or add a tracked address, all of the tokens from all other wallets disappear.

**Solution**:

In the make-tokens arm, when the token scry for the new address is null (it will be the first in the list of addresses), the trap just escapes rather than iterating through the rest of the addresses. I changed the ?~ to continue to the $ recursion.


dr-frmr:
Hey thanks. Sorry that I didn't see this until now. Looks great!

dr-frmr (approved):

PR #360, opened 2023-01-04
ziggurat: add %pyro-ships-ready update and scry

hosted-fornet:
**Problem**:

FE needs to be able to see what ships are running, according to %ziggurat.

**Solution**:

Add an update, scry path, and send updates for %pyro-ships-ready.

**Notes**:

FYI @0x70b1a5


tadad (approved):
nice, worked on my machine

PR #361, opened 2023-01-04
Da/ford cache

tadad:
**Problem**:

`:pyro|commit` misses the ford cache and is super slow

**Solution**:
turn all `page`s into `lobe`s, include the full pages in `lat.rang`

**Notes**:

also added some bug fixes in gen/pyro


hosted-fornet (approved):
I merged in next/pyro for testing: this works great, nice work!

It's annoying that such a small change to the `yoki` has such a big impact on performance, but well done tracking it down.

tadad:
got a recommendation from ted to do one big scry instead of N scries for N files. Going to try that tomorrow before merging - could save ~5secs per commit which would be extremely nice

hosted-fornet:
Would you mind merging this and then making a new PR with that change? I may want to build on this later this evening.

hosted-fornet:
That definitely sounds like a good change, though :+1:

lib/zig/pyro.hoon
```diff
@@ -60,6 +60,38 @@
   |=  [who=@p to=@p app=@tas =path]
   (poke who who %subscriber %subscriber-action [%sub to app path])
 ::
+++  park
+  |=  [=desk =beak]
```
hosted-fornet, line 64: Can you change the `beak` here to an `our` and `now`? Would be easier to use from Gall app that way.

hosted-fornet (changes requested):
`beak` -> `our` and `now` would make usage from gall much smooth

lib/zig/pyro.hoon
hosted-fornet: Another idea: maybe add another gate to the lib that calls `+park` and returns a `card` so the caller doesn't have to write their own.

hosted-fornet (commented):
+1 idea

lib/zig/pyro.hoon
hosted-fornet: > Can you change the beak here to an our and now? Would be easier to use from Gall app that way.

An alternative proposal: get rid of `desk` and use the [q.beak](https://github.com/urbit/urbit/blob/master/pkg/arvo/sys/lull.hoon#L805) to hold it. I leave it to your aesthetic judgement.

tadad: why would it be hard? `byk.bowl` is the beak

hosted-fornet: You are correct it is not hard. I think it would be cleaner to either specify args explicitly (perhaps in the order of a beak so it can be input as a beak from the generator?) or to use the `q.beak` to hold the `desk` and go all-in on it.

tadad: true

PR #365, opened 2023-01-05
misc. %pyro fixes

tadad:
**Problem**:
pyro code isn't clean
pyro needs more tooling
other misc fixes to run.hoon and ziggurat

**Solution**:
refactor it
comment it, remove superfluous comments
add more generators
add more helpers to lib/pyro
remove useless scries, refactor remaining ones

**Notes**:
@hosted-fornet only need a review on ziggurat.hoon and run.hoon


tadad:
@hosted-fornet do you want me to start working on swapping these subscriptions out in run/hoon thread?

ted/ziggurat/test/run.hoon
```diff
@@ -268,18 +267,13 @@
...
   ^-  form:m
   |-
-  ;<  is-next-events-empty=?  bind:m
-    (scry:strandio ? /gx/pyro/is-next-events-empty/noun)
   ;<  soonest-timer=(unit @da)  bind:m
     (scry:strandio (unit @da) /gx/pyre/soonest-timer/noun)
   ;<  now=@da  bind:m  get-time:strandio
-  ?.  is-next-events-empty
-    ;<  ~  bind:m  (wait:strandio (add now loop-duration))
-    $
```
tadad, line 282: `next-events` will always be empty because that's how `+plow-all` works

app/pyro.hoon
```diff
@@ -163,12 +150,6 @@
       [%fleet-ships t.t.path ~(key by u.sips)]
     ::
         [%x %pill ~]  ``pill+!>(pil)
-    ::
-        [%x %is-next-events-empty ~]
-      :^  ~  ~  %noun
-      !>   ^-  ?
-      %+  levy  ~(val by piers)
-      |=([p=pier] =(0 ~(wyt in next-events.p)))
```
tadad, line 171: it will always be empty

app/pyro.hoon
```diff
@@ -222,15 +187,15 @@
   ++  ahoy
     =?  fresh-piers  !(~(has by fresh-piers) who)
       %+  ~(put by fresh-piers)  who
-      [pier-data (~(get ja unix-boths) who)]
-    =/  =card:agent:gall  [%give %fact ~[/ready/(scot %p who)] %noun !>(%.y)]
-    ..ahoy:(emit-cards ~[card])
+      [pier-data(processing-events %.y) (~(get ja unix-boths) who)]
```
hosted-fornet, line 190: Is the reason for this change the switch in the default `processing-events` here?:

https://github.com/uqbar-dao/uqbar-core/blob/c5f070decd8da53b04ce0601de1da0dac9a04a02/app/pyro.hoon#L34

app/pyro.hoon
```diff
@@ -265,12 +229,12 @@
     ?:  =(~ next-events)
       ..abet-pe
     ?.  processing-events
-      ..abet-pe
+      ~&(%pyro^%not-plowing-events^who=who ..abet-pe)
     =^  ue  next-events  ~(get to next-events)
-    =/  poke-arm  (mox +23.snap)
+    =/  poke-arm  (mox +23.snap) :: see +poke in arvo.hoon
```
hosted-fornet, line 234: Good comment :+1:

sur/zig/pyro.hoon
```diff
@@ -40,38 +30,32 @@
...
+      ::  snapshot import/exports
+      ::
       [%export-snap =path]
       [%import-snap jam-file-path=path snap-label=path]
       [%export-fresh-piers ~]
       [%import-fresh-piers jam-file-path=path]
-      [%clear-snaps ~]
-      ::  pill
-      ::
-      [%pill =pill]
```
hosted-fornet, line 51: Just noting here that removing %pill from %pyro-action means we will have to write its own mark file if we wish to make this accessible from FE. For now, this is fine since I don't think we have plans to make this accessible from FE (something we should either update periodically in OTAs or a bit of a poweruser feature).

lib/zig/pyro.hoon
```diff
@@ -33,6 +51,25 @@
...
+        ::
+          %+  lien  p.q.uf
+          |=  =blit:dill
+          ?.  ?=(%lin -.blit)
+            |
+          !=(~ (find tape p.blit))
+      ==
+    (pure:m ~)
+  $
+::
```
hosted-fornet, line 72: Just noting here that we should check in sometime down the road to see if we've made use of these gates.

hosted-fornet (changes requested):
A couple of things I'd like addressed -- otherwise just a couple questions and notes to our future selves.

To explain a bit more about my style comments: my general policy is that when modifying existing code, I try to follow the style in that piece of code. Ideally, a whole repo has the same style -- but this is hard to maintain without a lot of discipline and agreement amongst contributors and I don't think its a reasonable goal for Uqbar right now. Since we don't have a code style guide and are moving fast, I just aim to follow whatever file I am in.

hosted-fornet:
FYI @0x70b1a5 a number of changes here to %pyro scry interface, so if you are using %pyro scries, you should take a look at this PR.

https://github.com/uqbar-dao/uqbar-core/blob/d1727d0925a780659e2b6ccb5c1b6028beb5009d/sur/zig/pyro.hoon#L51-L56

https://github.com/uqbar-dao/uqbar-core/blob/d1727d0925a780659e2b6ccb5c1b6028beb5009d/app/pyro.hoon#L107-L124

PR #366, opened 2023-01-05
ziggurat: add %edit-test

hosted-fornet:
**Problem**:

Current way to "edit" a test is to %delete-test and then %add-test, but this is not easy for FE.

**Solution**:

Add an %edit-test poke.

**Notes**:

None


tadad (approved):
worked on my machine

PR #367, opened 2023-01-06
ziggurat: add configs state

hosted-fornet:
Problem:

Want to be able to accept configurations for services beyond just %sequencer.

Solution:

Change some state (virtualnet-addresses and town-sequencers) to a more general configs which is placed into test-steps test-globals on run.

Notes:

This is setting the stage for service ships and configuration files. Next I will work on pyro:ziggurat file stuff and service ships and configuration files.

Replaces PR #363


hosted-fornet:
Oh shoot. I meant beb5b89 to go to a new PR.

hosted-fornet:
Ok fixed. :sweat_smile:

zig/custom-step-definitions/send-wallet-transaction.hoon
```diff
@@ -7,7 +7,9 @@
           expected=(list test-read-step:zig)
       ==
   ^-  test-steps:zig
-  =/  address=@ux  (~(got by addresses:test-globals) who)
+  =/  address=@ux
+    %.   [who %address]
+    ~(got by (~(got by configs:test-globals) ''))
```
tadad, line 12: why are you fetching `''` as the project name?

tadad: shouldn't the project name be passed in as an argument to the test?

readme-pyro.md
```diff
@@ -185,10 +185,15 @@ Examples can be seen in the `zig/test-steps/` dir](https://github.com/uqbar-dao/
 
 The subject of a `test-steps` is defined by the `/=` imports at the top of the `test-steps` file.
 In addition, this subject will be applied for [`custom-step-definitions`](#custom-inputs), so those dependencies must be included in `test-steps`.
-Finally, some `test-globals` will be accessible by `test-steps`.
-`addresses:test-globals` is the `(map @p @ux)` defined in `%ziggurat` app state and set with the `%set-virtualnet-addresses` action.
-The `addresses:test-globals` map is useful for easy access and pairing between virtualships and their testnet addresses.
-`test-results:` are also accessible, so that the results of a previous `test-step` is usable in the current one (TODO).
+Finally, some `test-globals` will be accessible by `test-steps` (see sur/zig/ziggurat.hoon):
+
+`configs:test-globals` is a `(map [project-name=(unit @t) who=@p what=@tas] @)` that stores general data that can be added to with `%add-config`.
```
tadad, line 190: `(mip project-name=@t [who=@p what=@tas] @)`

app/ziggurat.hoon
```diff
@@ -51,8 +51,11 @@
     :*  %0
         ~
     ::
-        %-  ~(gas by *(map @p address:smart))
-        ~[[~nec nec-address] [~bud bud-address]]
+        %+  ~(put by *configs:zig)  ''
+        %-  ~(gas by *config:zig)
+        :+  [[~nec %address] nec-address]
+          [[~bud %address] bud-address]
+        ~
```
tadad, line 58: oh I see you are putting the these in project-name `''`. Is that a good idea? Something about this feels a little unclean

hosted-fornet: > oh I see you are putting the these in project-name ''

Correct. The idea here is that there are project-specific pieces of state that we want to store as well as global pieces of state.

One alternative possibility I can think of is making the outer `configs` a `(map (unit @t) config:zig)`. Does that seem cleaner?

readme-pyro.md
hosted-fornet: Good catch, this didn't get updated. I may go ahead and change the map-of-maps to mips for ease of access while I'm at it.

app/ziggurat.hoon
hosted-fornet: Ping @tadad ; any thoughts here?

Another possibility is reserving the `'global'` key for global pieces of state. Open to other suggestions as well.

tadad: I like `'global'`, seems clean enough

PR #368, opened 2023-01-06
ziggurat: add configs and integrate %pyro file syncing

hosted-fornet:
**Problem**:

1. We want to really actually sync files from devship to virtualships.
2. We want to make projects importable: point to a desk and bring it into %ziggurat.
3. We want to have configuration files such that when a desk is brought in it can set %ziggurat state, sync to specific virtualships, and set the project in a specific state.

**Solution**:

1. Make use of %pyro `+park` to keep files synced b/w devship and virtualships, as dictated by `sync-desk-to-vship`.

2&3. Modify %new-project. Add configuration files (see, e.g., `/zig/configs/zig.hoon`). Add `+load-config` to load them. Add `+cis` (commit-install-start) core to handle walking through the desired configuration steps.

**Notes**:

Still TODO as of a8a6754:
* [x] Increase robustness
* [x] Output current project configuration (as determined by state) to a new configuration file
* [x] Better error handling and output
* [x] Remove debugging `~&`s


tadad (approved):
looks good

PR #371, opened 2023-01-09
ziggurat: change test-steps enjs format

hosted-fornet:
**Problem**:

JSON HTTP interface of %ziggurat is suboptimal.

**Solution**:

Change nesting of `test-steps` and `update`. Change all underscores to dashes for consistency.

**Notes**:

FYI @tadad

PR #372, opened 2023-01-10
ziggurat: test steps json tweaks tweak

hosted-fornet:
**Problem**:

Changes were not as good as we thought on PR #371

**Solution**:

Revert #371 and apply the stuff we actually want here.

**Notes**:

Not sure if this will actually work because github might get too fancy with the merging; if not I'll rebase and force-push to next/pyro directly.
FYI @0x70b1a5


hosted-fornet:
Yeah this isn't going to work. Force push to next/pyro inc.

PR #374, opened 2023-01-11
ziggurat: add `%test-queue` update

hosted-fornet:
**Problem**:

FE does not know what the state of the queue is.

**Solution**:

Send a `%test-queue` update to the FE about the queue whenever it is updated. Also add a `/test-queue` scry endpoint.

**Notes**:

FYI @tadad 

@0x70b1a5 let me know if you prefer these queue updates to not be associated with a project and I can make that change. Then you would subscribe, e.g., to `/project/''` and all of these `%test-queue` updates would come to that endpoint.


tadad (approved):
nice

PR #376, opened 2023-01-12
ziggurat: add /pyro-agent-state/[who]/[app] scry

hosted-fornet:
**Problem**:

It is painful to use the dbug-dashboard, but we need to get state of virtualship agents to FE somehow.

**Solution**:

Scry from dev ship into %subscriber running on virtualship. %subscriber then scries agent /dbug/state, and turns that into a cord, before returning it. Since we know it is a cord coming back, we can `;;` it and pass to FE.

**Notes**:

FYI @tadad


hosted-fornet:
I tested by scrying from within Urbit as well as

```
export NEC_COOKIE=$(curl -i -X POST localhost:8080/~/login -d 'password=ropnys-batwyd-nossyt-mapwet' | grep set-cookie | awk '{print $2}' | awk -F ';' '{print $1}')
curl --cookie "$NEC_COOKIE" localhost:8080/~/scry/ziggurat/pyro-agent-state/~nec/wallet.json  | jq
curl --cookie "$NEC_COOKIE" localhost:8080/~/scry/ziggurat/pyro-agent-state/~nec/indexer.json  | jq
```

Note that the result will be pretty-printed UNLESS it is too big (big data takes VERY LONG to pretty print). I chose the cutoff somewhat arbitrarily, so we may want to alter it or make it configurable:

https://github.com/uqbar-dao/uqbar-core/blob/6335895f3f067a2caf4676725981747495f2e7e8/app/subscriber.hoon#L86

app/ziggurat.hoon
```diff
@@ -1221,50 +1145,25 @@
...
-      %-  dashboard-error
-      %-  crip
-      ;:  weld
-          "app {<app>} subject failed to build; fix sur"
-          " file path and re-add with %add-app-to-dashboard."
-          " error message from build: {<p.mold.u.dbug>}"
-      ==
+      [%pyro-agent-state @ @ ~]
+    =*  who      i.t.t.p
+    =*  app      `@tas`i.t.t.t.p
```
0x70b1a5, line 1150: shouldn't it be `i.t.p` and `i.t.t.p` ?

0x70b1a5 (changes requested):
just one thing i noticed with the path.

app/ziggurat.hoon
hosted-fornet: Good eye. What is confusing here is we are switching on `+.p`, see:

https://github.com/uqbar-dao/uqbar-core/blob/6335895f3f067a2caf4676725981747495f2e7e8/app/ziggurat.hoon#L1071

and `p` here is actually something like

/x/pyro-agent-state/~nec/foo

0x70b1a5 (approved):

PR #380, opened 2023-01-17
ziggurat: add test-id to test updates

hosted-fornet:
**Problem**:

The FE has to work hard to store `test-id` for updates.

**Solution**:

Just return the `test-id` as a field, in addition to the object key.

**Notes**:

LMK if this is sufficient or if other things need to have a `test_id` field as well.

FYI @tadad


0x70b1a5 (approved):
LGTM :+1:

PR #388, opened 2023-01-19
read desks scry

0x70b1a5:
**Problem**:

%ziggurat needs to know which desks a user has in order to be able to inform them that creating a new project with an identical name to an existing desk will result in an attempt to import that desk as a ziggurat project. 

**Solution**:

Use this on-peek to scry Clay for desk information, so that FE can surface it to user.

**Notes**:

🔍 🦧


hosted-fornet (approved):
LGTM :+1:

PR #390, opened 2023-01-20
ziggurat: improve %add-sync-desk-vships

hosted-fornet:
**Problem**:

FE can't easily commit and install for new ships.

**Solution**:

Improve %add-sync-desk-vships to use `+cis` door machinery and give out `update:zig`s.
Add %send-pyro-dojo poke.

**Notes**:

The `%add-sync-desk-vships` poke now takes additional arguments:
```hoon
[project=@t request-id=(unit @t) %add-sync-desk-vships ships=(list @p) install=? start-apps=(list @tas)
```

If ships are provided but install is `%.n`, the only thing that will happen is the desk will be synced from devship to vship and will be registered to be kept in sync. This is regardless of `start-apps`.

If ships are provided and install is `%.y`, `|install our %foo` will be run on the virtualships after the syncing is done.

If ships are provided and install is `%.y` and there are apps in `start-apps`, these apps will be started like `|start %foo %my-app`. This is not necessary if your app is on desk.bill: those will be started during installation. This is for additional apps contained in the desk but not in desk.bill.

FYI @tadad


0x70b1a5 (approved):
Looks Good to Monkey 🔍 🦧

PR #397, opened 2023-01-26
Da/commit

tadad:
**Problem**:
|commits and |install s take O(n) time

**Solution**:
Utilize the new caching to make it O(1)ish time

**Notes**:
speed issues are fixed by [this pr](https://github.com/urbit/urbit/pull/6254)


hosted-fornet:
This is looking good. I'd suggest changing the API for commits/installs that share cache. It would be nice from a user's perspective if I could just call `commit:pyro-lib` or `install:pyro-lib` for a single ship and never have to worry about caching and %pyro just takes care of it for me behind the scenes: if there is a cache that matches the desk I am committing or installing on another ship, it is used; if not, commit/install from scratch.

E.g., if I spin up a new ship, it is more natural for me to call a `commit` and/or `install` gate then to go run a thread where I have to put in arguments that include both an already cached ship as well as my new ship.

It would also give more flexibility for designing systems that layer on top of %pyro. E.g., if I am setting up multiple ships across multiple operations (say committing, installing, starting apps not in desk.bill, running a series of pokes, etc to get each ship into a useful testing state), it is nicer for me to have the option of either:

1. Commit all ships; then install on all ships; then start apps on all ships; then run series of pokes on all ships
2. Run one separate "process" for each ship where each process is sequentially doing commit, install, start, run series of pokes

The API exposed here with the threads limits me, making 2 pretty unnatural.

tadad:
good point - just refactored it so that every time you `apex`/`abet-pe` a ship, the cache gets `uni by`d and refilled. So now running `:pyro|commit` twice will automatically fill the cache

PR #400, opened 2023-01-30
Sm/try to make pe convert

0x70b1a5:
**Problem**:

Pyro vships don't have a `grow` for dojo `%blit`s, so the frontend cannot receive updates on dojo display.

**Solution**:

Create the `grow`.


lib/zig/dill.hoon
```diff
@@ -0,0 +1,94 @@
+::  dill: utilities for dill's data structures
```
hosted-fornet, line 1: Would prefer this as a symlink to `../../../base-dev/lib/dill.hoon` so we are just pointing at that as a source of truth. Assuming you've set up your uqbar-core as an urbit repo submodule following directions in readme https://github.com/uqbar-dao/uqbar-core#initial-installation you can do this with something like

```bash
cd lib/zig
ln -s ../../../base-dev/lib/dill.hoon .
```

lib/zig/pyro.hoon
```diff
@@ -185,6 +185,21 @@
     :-  %a
     %+  turn  paths
     |=(p=^path (path p))
+  ::
+  ++  pyro-effect
+    |=  [who=@p ufs=unix-effect]
+    ^-  json
+    ?+    -.+.ufs  ~  :: ignore non-%blit
```
hosted-fornet, line 192: Prefer avoiding using `-` or `+` when possible -- thought in some cases like tagged unions you do have to use `-`. So here, would prefer something like

```hoon
?+    -.q.ufs  ~  :: ignore non-%blit
```

since a `pair` has head with `p` face and tail with `q` face.

lib/zig/pyro.hoon
```diff
@@ -185,6 +185,21 @@
...
+        %blit
+      %-  pairs
+      :~  [%ship s+(scot %p who)]
+        :-  %blits
+        :-  %a  
+        %+  turn  `(list blit:dill)`+.+.ufs
```
hosted-fornet, line 198: Similar to above, would prefer `p.q.ufs`

hosted-fornet: Since the `p` in the tagged union for %blit refers to the thing we want: 
https://github.com/uqbar-dao/dev-suite/blob/master/sur/zig/pyro.hoon#L34

lib/zig/pyro.hoon
```diff
@@ -1,6 +1,6 @@
 /-  *zig-pyro,
     spider
-/+  *strandio
+/+  *strandio, zig-dill
```
hosted-fornet, line 3: Nit: this isn't the case for all our files, but here we have convention of one import per line, so would prefer

```hoon
/+  *strandio,
    zig-dill
```

hosted-fornet (approved):
After changes, LGTM.

In the future, please make PRs to https://github.com/uqbar-dao/dev-suite
We are still in a bit of a transition period, and since this PR is already here, I'll just copy the changes over once it is merged.

0x70b1a5:
@hosted-fornet merged... sorry about wrong target!

PR #401, opened 2023-01-30
nuked

tadad:
**Problem**:
separating dev-suite and uqbar-core

**Solution**:
delete everything

**Notes**:
cleaning up some other stuff while I see it


tadad:
@hosted-fornet let me know if you want to delete ziggurat here or on another branch

hosted-fornet:
> @hosted-fornet let me know if you want to delete ziggurat here or on another branch

@tadad Sorry I didn't see this till now: yes, please do delete %ziggurat here.

mar/indexer/headers.hoon
```diff
@@ -1,5 +1,5 @@
 /-  ui=zig-indexer,
-    zig=zig-ziggurat
+    zig=zig-ziggurat :: TODO
```
tadad, line 2: @hosted-fornet - not sure what to do here? an indexer mark shouldn't really have a ziggurat dependency, but I'll let you sort this out

tadad:
also now that I think about it, our desk.docket-0 might need to change as well? And the frontend obviously though maybe @0x70b1a5 is already on it?

PR #419, opened 2023-03-01
update glob 0.0.11

0x70b1a5:
**Problem**:

We don't have glob 11, including all sorts of pheachures and uppedaights.

**Solution**:

Globbe.

**Notes**:

DOWNLOAD AT OWN RISK.


dr-frmr (changes requested):
Sorry -- can you change the `version` to `[0 1 1]`? I forgot to commit that, but the current version is 0.1.0

PR #424, opened 2023-03-23
wallet: add tx-hash to receipt

bitful-pannul:
**Problem**:

Wallet doesn't pass on the tx-hash to app origin

**Solution**:

Include the hash in the wallet-update type

**Notes**:
.

### DIFF
--- a/app/wallet.hoon
+++ b/app/wallet.hoon
@@ -182,7 +182,7 @@
         %+  ~(poke pass:io /receipt)
           [our.bowl p.u.origin.u.found]
         :-  %wallet-update
-        !>(`wallet-update`[%sequencer-receipt origin.u.found +.q.result])
+        !>(`wallet-update`[%sequencer-receipt origin.u.found p.result +.q.result])
       ==
     =^  cards  tokens
       ?.  ?=(%receipt -.q.result)  [cards tokens]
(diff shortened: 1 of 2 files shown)


app/wallet.hoon
dr-frmr, line 1: looks good

sur/zig/wallet.hoon
```diff
@@ -108,7 +108,7 @@
           action=supported-actions
       ==
       ::  poked back to origin after sequencer optimistically processes
-      [%sequencer-receipt =origin sequencer-receipt:uqbar]
+      [%sequencer-receipt =origin tx-hash=id:smart sequencer-receipt:uqbar]
```
dr-frmr, line 111: just for consistency's sake, can we do `=hash:smart` here?

dr-frmr (changes requested):

dr-frmr:
I will merge this for `release-0.1.3` after the face change!

dr-frmr:
Unfortunately, since other apps (pokur and pongo) use the `wallet-update` mold, this is a breaking change and we'll have to coordinate the update so that they don't fail upon receiving receipts. Can make it work for this update since we control distribution for those apps, but in the future, won't be able to change the interface like this without making a new mold!

dr-frmr (approved):
LGTM, merging

PR #435, opened 2023-04-10
ziggurat: clean up old %ziggurat stuff

hosted-fornet:
**Problem**:

%ziggurat project setup files now live in https://github.com/uqbar-dao/zig-dev but still have deps here.

**Solution**:

Remove deps.

**Notes**:

None


dr-frmr:
beautiful diff

dr-frmr (approved):

PR #442, opened 2023-04-20
rollup, sequencer, indexer: convert subscription-based comms to pokes

dr-frmr:
**Problem**:

Subscriptions between sequencer, rollup agent, and indexers are buggy. Ship->ship subscriptions aren't good about re-establishing themselves and have led to looping bugs a few times.

**Solution**:

Use a poke-only model for sharing batches and rollup governance data.

**Notes**:

Did some major refactoring here


dr-frmr:
This is ready

dr-frmr:
Next step: test transition between old subscription model and this

PR #444, opened 2023-04-26
hotfix: engine: skip +charge on failed gas audit

dr-frmr:
**Problem**:

Execution engine refactor made it so that +charge arm was not getting properly skipped if a transaction failed gas audit

**Solution**:

Check errorcode on transaction output before calling charge

**Notes**:

Deployed immediately to testnet


bitful-pannul (approved):
LGTM

PR #447, opened 2023-05-02
sequencer: post batches to rollup contract

dr-frmr:
**Problem**:

Sequencer needs to post batches to rollup to validate and perform data availability. Time to sunset the "rollup simulator" app and start using the real rollup contract!

**Solution**:

Solution is in two parts:
- a poke for rollup-tracking sidecar to share new calculated town root after batch posting
- a scry path for sequencer app to share completed batch with sidecar

The old rollup simulator will also be removed from the codebase now, along with the aspects of the sequencer and indexer apps that rely on it.

**Notes**:

This PR also updates uqbar-core to %zuse kelvin 413.


dr-frmr:
This is ready for next release

PR #450, opened 2023-05-08
ueth deposit hardcode removed

tadad:
**Problem**:
Deposits are hardcoded to ueth only

**Solution**:
Make it general for any ERC20/Fungible pair

**Notes**:


app/sequencer.hoon
```diff
@@ -222,31 +221,36 @@
         ?~  working-batch
           chain.u.town
         chain.u.working-batch
+      =/  =data:smart
+        =+  got=(get:big p.working-chain token.deposit)
+        ?~  got  ~|("%sequencer: token doesn't exist" !!)
```
dr-frmr, line 226: We need to handle this case, cause this is what happens if someone deposits a "new" token, AKA a token that has never been deposited into the rollup before

PR #451, opened 2023-05-10
Da/arbitrary bridging

tadad:
**Problem**:
Token bridging is only for uETH, and also lives in the sequencer

**Solution**:
Move token bridging to the engine, let it accept deposits from all L1 tokens

**Notes**:


sur/zig/sequencer.hoon
```diff
@@ -68,23 +68,13 @@
       [%del-block-height-api-key ~]
       [%clear-state ~]
       ::  transactions
-      [%deposit deposit-bytes=tape]  ::  from sidecar
```
dr-frmr, line 71: You will still need a poke here for the sequencer app to ingest deposit-bytes from the sidecar

lib/zig/sys/engine.hoon
```diff
@@ -10,14 +10,95 @@
...
+          `@ux`(scan `tape`(slag 2 i.deposits) hex)
+        =/  =deposit
+          =+  (rev 3 [192 deposit-bytes])
+          :*  town-id=(rev 3 32 (end [3 32] -))
+              deposit-index=(rev 3 32 (cut 3 [32 32] -))
+              token=(rev 3 32 (cut 3 [64 32] -))
+              destination-address=(rev 3 32 (cut 3 [96 32] -))
+              amount=(rev 3 32 (cut 3 [128 32] -))
+              block-number=(rev 3 32 (cut 3 [160 32] -))
+          ==
```
dr-frmr, line 36: Makes sense to me to perform this in the sequencer app on-poke then pass in a list of `deposit`s to the engine instead of a list of tapes

lib/zig/sys/engine.hoon
```diff
@@ -10,14 +10,95 @@
...
+            =+  ;;(token-metadata noun.p.u.item)
+            u.item(noun.p -(supply (add supply.- amount.deposit)))
+          :*  %& 
+              metadata-id
+              bridge-contract-id:smart
+              0x0
+              town-id
+              token.deposit
+              %token-metadata
+              :: TODO the token metadata is very fucked, double check this before merging
```
dr-frmr, line 63: Need to change what the sidecar gives us to grab these things (sidecar should make RPC call and include metadata values in the %deposit poke imo)

lib/zig/sys/smart.hoon
```diff
@@ -78,6 +78,8 @@
 ++  zigs-contract-id  `@ux`'zigs-contract'
 ::  hardcoded contract for ETH deposited on Uqbar
 ++  ueth-contract-id  `@ux`'ueth-contract'
+::  hardcoded contract for all bridged fungible tokens
+++  bridge-contract-id  `@ux`'bridge-contract'
```
dr-frmr, line 82: If you change smart.hoon we're back in breaking changes world -- try keeping this in engine.hoon for now

dr-frmr:
Left some comments, I think we should make an engine library file to put helper functions (such as the ones at the bottom of the file) so that the blob of code that handles this isn't such a large % of the LoC of the main engine

lib/zig/sys/engine.hoon
tadad: I think we should change it to just a @ux instead of a tape, and store the @ux es in sequencer. We can keep this chunk of code in engine, and that way we don't have to do any `+on-load` migration stuff in sequencer

tadad: see latest code, I think it works quite well

dr-frmr (approved):

PR #455, opened 2023-05-15
Da/nft bridging

tadad:
**Problem**:
NFT bridging isn't enabled

**Solution**:
Enable it in the engine

**Notes**:
some extra fixes and logic to fungible deposits as well


lib/zig/sys/engine.hoon
```diff
@@ -1,5 +1,20 @@
...
+          symbol=@t
+          properties=(pset:smart @tas)
+          supply=@ud
+          cap=(unit @ud)  ::  (~ if no cap)
+          mintable=?      ::  automatically set to %.n if supply == cap
+          minters=(pset:smart address:smart)
+          deployer=id:smart
+          salt=@
+      ==
+    --
```
tadad, line 16: should I import /con/compiled/nft/hoon for this mold? Or just leave it her

dr-frmr:
Please build off this PR: https://github.com/uqbar-dao/uqbar-core/pull/454

lib/zig/sys/engine.hoon
dr-frmr: yeah ideally import `/con/lib/nft/hoon`

con/lib/nft.hoon
```diff
@@ -180,6 +180,8 @@
   ++  deploy
     |=  [=context act=deploy:sur]
     ^-  (quip call diff)
+    ::  make salt unique by including deployer + their input
+    =/  salt  (cat 3 salt.act id.caller.context)
```
tadad, line 184: if this isn't the case, then someone can reserve the salt for an L1 nft collection, and get the engine into an inconsistent state when someone tries to bridge that L1 nft collection to uqbar

dr-frmr (approved):

Issue #460, opened 2023-05-22
`/lib/zink/conq.hoon` is broken


tadad:
I believe this happened when the zk tools were removed from zink?

PR #461, opened 2023-05-22
conq fixed

tadad:
**Problem**:
see #460 : `conq.hoon` broke when zk tools like pedersen were removed from uqbar-core. This means that we can't call `+zig!compile`

**Solution**:
remove unnecessary zk tooling in `conq.hoon`.

**Notes**:
let me know if there is anything worth saving that I deleted


dr-frmr:
just saw this, sorry yeah realized and deleted the bad code

PR #467, opened 2023-06-01
wallet: add %unsigned-transaction poke

bitful-pannul:
With account abstraction, we can have a contract pay gas and not sign or worry about signatures. Add a poke for it in wallet so other apps can use it and get back receipts. 

Could bake into the normal %transaction method and put in pending-store, but not pretty. 

Small error in on-agent jab py transaction store.

### DIFF
--- a/app/wallet.hoon
+++ b/app/wallet.hoon
@@ -541,6 +541,43 @@
         pending-store  (~(put by pending-store) from.act my-pending)
       ==
     ::
+        %unsigned-transaction
+      =/  =caller:smart
+        :+  0x0
+          0
+        :: zigs account of contract?
+        (hash-data:engine zigs-contract-id:smart contract.act town.act `@`'zigs')
+      ::  build calldata of transaction, depending on argument type
+      =/  =calldata:smart
+          ;;(calldata:smart +.action.act)
+      ::  
+      =/  =shell:smart
+        :*  caller
+            eth-hash=~
+            to=contract.act
+            gas=gas.act
+            town.act
+            status=%101
+        ==
+      ::  generate hash
+      =/  hash  (hash-transaction [calldata shell])
+      =/  tx=transaction:smart  [[0 0 0] calldata shell]
+      ~&  >>  "%wallet: submitting unsigned tx with hash {<hash>}"
+      ::  update stores
+      :_  %=    state
+              unfinished-transaction-store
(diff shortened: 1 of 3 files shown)


app/wallet.hoon
```diff
@@ -540,6 +540,43 @@
       %=  state
         pending-store  (~(put by pending-store) from.act my-pending)
       ==
+    ::
+        %unsigned-transaction
+      =/  =caller:smart
+        :+  0x0
+          0
+        :: zigs account of contract?
+        (hash-data:engine zigs-contract-id:smart contract.act town.act `@`'zigs')
```
dr-frmr, line 549: this is unnecessary -- can change this whole `:+` into `[0x0 0 0x0]` i think

app/wallet.hoon
```diff
@@ -540,6 +540,43 @@
...
+      ::  build calldata of transaction, depending on argument type
+      =/  =calldata:smart
+          ;;(calldata:smart +.action.act)
```
dr-frmr, line 552: no need to `;;` here?

app/wallet.hoon
```diff
@@ -540,6 +540,43 @@
...
+      ::  
+      =/  =shell:smart
+        :*  caller
+            eth-hash=~
+            to=contract.act
+            gas=gas.act
+            town.act
+            status=%101
```
dr-frmr, line 560: why initialize with status `101` vs. `100` here? also, since this is so smol, can probs just do `[caller ~ contract.act gas.act town.act %100]` on one line

app/wallet.hoon
```diff
@@ -540,6 +540,43 @@
...
+              hash
+            [origin.act tx action.act ~]
+          ==
+      :~  (tx-update-card hash tx action.act)
+          :*  %pass  /submit-tx/(scot %ux hash)
+              %agent  [our.bowl %uqbar]
+              %poke  %uqbar-write
+              !>(`write:uqbar`[%submit tx])
+          ==
+      ==
```
dr-frmr, line 579: consolidate `==` onto same line as inner one

dr-frmr (changes requested):
see comments, and, it would be good to add parsing for this action type in lib/zig/wallet.hoon so that the frontend can eventually use it

dr-frmr (approved):

PR #468, opened 2023-06-02
batcher-interval: get compiling again

hosted-fornet:
**Problem**:

%batcher-interval doesn't compile -- which interferes with dev-suite installation of master since it compiles all apps.

**Solution**:

Fix bugs.

**Notes**:

None


dr-frmr (approved):

dr-frmr:
Sorry about that!

PR #472, opened 2023-06-07
nft contract: add push+pull, fix bugs, write tests

dr-frmr:


bitful-pannul:
A note on %pull, if we add signatures we need to store a nonces map in the nft noun

bitful-pannul (approved):

PR #480, opened 2023-06-30
engine: reproduce & validate eth-hashes for transactions

bitful-pannul:
when a tx gets signed from an external eth wallet, hash and verify it in engine too. 

this awaits some changes in UI to clean up the string that gets signed.
related with signed-type-messages: https://github.com/uqbar-dao/uqbar-core/pull/478

add tests to lib/engine [x]

### DIFF
--- a/app/wallet.hoon
+++ b/app/wallet.hoon
@@ -485,10 +485,10 @@
       ::  take in a new pending transaction
       =/  =caller:smart
         :+  from.act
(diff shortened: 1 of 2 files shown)


bitful-pannul:
note nonce change in app/wallet, if we leave it ephemeral, there's no way for the engine to reproduce that entropy. could also add it to the transaction type..

